Issue #26, opened 2020-05-15
Improve sub application auto deployment


filippomc:
Sub applications shoud be defined and behave exactly as root applications.
An sub application is any key in the values files with an `harness` key on it.
Auto deployment, service, etc, should work for sub applications exactly as it happens for the main application.
Use a recursive strategy to support `harness` key at any level.

filippomc:
some functionality is already present, needs to be tested and documented

PR #27, opened 2020-05-15
Feature/25: Improve harness auto configuration

filippomc:


utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-services.yaml
```diff
@@ -3,26 +3,25 @@
...
   ports:
-    - port: {{ .app.port }}
+    - port: {{ .app.harness.deployment.port }}
       name: http
 {{- end }}
 
 
 {{- range $app := .Values.apps }}
-  {{- if and (hasKey $app "port") ($app.autoservice | default true)  }}
+  {{- if and (hasKey $app "port") $app.autoservice  }}
```
zsinnema, line 20: is this correct? port is in the harness part of app

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-services.yaml
```diff
@@ -3,26 +3,25 @@
...
 ---
-  {{ include "deploy_utils.service" (dict "root" $ "app" $app) }}
   {{- range $subapp := $app }}
   {{- if contains "map" (typeOf $subapp)  }}
-  {{- if and (hasKey $subapp "port") ($subapp.autoservice | default false) }}
+  {{- if and (hasKey $subapp "port") $subapp.autoservice }}
```
zsinnema, line 24: same here

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-services.yaml
filippomc: You're right. This is wrong but still working because old values are kept

zsinnema: I think we should also add autoservice to the template with a default false

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-services.yaml
filippomc: Addressed. As a reminder  subapps have to be tested and possibly reimplemented (now 1 level, a recursive strategy would fit better) - see #26

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-services.yaml
filippomc: Addressed

Issue #33, opened 2020-06-10
Create Python api to retrieve infrastructure information


filippomc:
The idea behind this is to give some applications the capability to make configurations depending on other applications in the cluster.

Information comes from the harness section of the values file. Example
```yaml
apps:
  workflows:
    harness:
      name: workflows
      subdomain: workflows
      domain: null
      secured: false
      deployment:
        auto: false
        replicas: 1
        image: localhost:5000/workflows:2
        name: workflows
        port: 8081
        resources: &id001
          requests:
            memory: 32Mi
            cpu: 25m
          limits:
            memory: 500Mi
            cpu: 500m
      service:
        auto: true
        name: workflows
        port: 8080
      deploy:
        auto: true
        port: 8080
    image: localhost:5000/workflows:2
    name: workflows
    port: 8081
    resources: *id001
```

Sample usage
```python
from harness import applications
app_conf = applications.get_configurations()
for app in app_conf:
    if app['deployment']['auto']:
        # DO SOMETHING
```

The values.yaml file can be injected in a configmap like this:
```yaml
kind: ConfigMap
apiVersion: v1
data:
  allvalues.yaml: |
    {{- .Values | toYaml | nindent 4 }}
```
This configmap can be used inside auto-deployments.yaml to be injected inside every automatic deployment.

zsinnema:
I've changed the interface a bit to have a more convenient interface for accessing apps
```
from cloudharness.utils.config import conf
conf.get_application_by_filter(harness__deployment__auto=True) # get all auto deploy apps
conf.get_application_by_filter(name='workflows') # get the workflows app
``

PR #35, opened 2020-06-12
Feature/33 Create Python api to retrieve infrastructure information

zsinnema:
Create Python api to retrieve infrastructure information
    
    see #33 

added conf object to cloudharness utils, reworked env functions to use the conf object
mount helm values.yaml to autodeploy applications


libraries/cloudharness-common/cloudharness/utils/config.py
```diff
@@ -0,0 +1,49 @@
+import yaml
+
+ALLVALUES_PATH = '/opt/cloudharness/resources/allvalues.yaml'
+
+class _obj(object):
```
filippomc, line 5: Class should be uppercase (PEP8). A more informative name also would help read the code (same below). As for the package private, I prefer the convention to importwhat it considered api in the main package __init__ file rather than these clunky _ names

libraries/cloudharness-common/cloudharness/utils/env.py
```diff
@@ -79,21 +89,21 @@ def name_to_variable(application_name):
 
 # CloudHarness Events
 def get_cloudharness_events_client_id():
-    return get_variable('CH_KEYCLOAK_WEBCLIENT_ID')
+    return 'web-client'
```
filippomc, line 92: The same value is available from `.Values.apps.accounts.webclient.id`

libraries/cloudharness-common/cloudharness/utils/env.py
```diff
@@ -4,28 +4,38 @@
...
 
 HERE = os.path.dirname(os.path.realpath(__file__))
 
 
 def set_default_environment():
-    with open(HERE + '/resources/values.yaml') as f:
-        values = yaml.safe_load(f)
-        os.environ.update({v['name']: str(v['value']) for v in values['env'] if v['name'] not in os.environ})
+    values = conf.allvalues
+    os.environ.update({v['name']: str(v["value"]) for v in values['env'] if v['name'] not in os.environ})
```
filippomc, line 25: I don't really understand the need of this: wouldn't it be easier to get a value from the conf itself rather than creating env variables then looking for the env variable?

utilities/cloudharness_utilities/deployment-configuration/helm/templates/configmap.yaml
```diff
@@ -0,0 +1,22 @@
+{{- /* Be careful with json typos here, mainly trailing ',' */}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.apps.accounts.name | quote }}
```
filippomc, line 5: Why an application-specific config map here?

libraries/cloudharness-common/cloudharness/utils/env.py
zsinnema: yes agree

libraries/cloudharness-common/cloudharness/utils/config.py
zsinnema: the _ names indicate that these classes are "private"
about class name, will change it

utilities/cloudharness_utilities/deployment-configuration/helm/templates/configmap.yaml
zsinnema: ow yes agree, this is a copy&paste error, will remove it

libraries/cloudharness-common/cloudharness/utils/config.py
filippomc: I understand these are package private (see comment above), I just don't see any reason to make these private. CloudHarnessConfig surely shouldn't be as its interface is important to be known outside and may be used in type hinting. So my suggestion is to use the export at the module top level to make the module api explicit rather than make everything private.

zsinnema:
> Thanks @zsinnema , a few questions/change requests:
> 
> * is bootstrap.sh.tar here on purpose?
> * the api requested in the card is not implemented
> * some previous comments have not been addressed/answered

bootstrap.sh.tar removed, on accident committed. may be we add an issue to remove the files after it has been run/used

api is almost the same, only some other naming
`
from cloudharness.utils.config import conf
app_conf = conf.get_applications()
for app in app_conf:
    if app['deployment']['auto']:
        # DO SOMETHING
`
about comments:
can you please be a bit more specific about this?

filippomc:
> > Thanks @zsinnema , a few questions/change requests:
> > 
> > * is bootstrap.sh.tar here on purpose?
> > * the api requested in the card is not implemented
> > * some previous comments have not been addressed/answered
> 
> bootstrap.sh.tar removed, on accident committed. may be we add an issue to remove the files after it has been run/used
> 
> api is almost the same, only some other naming
> 

Almost the same is not same! We need some control especially in the common library, different people contribute here and different project can use it so the style must be kept consistent and issues should be a source of truth for the project history.

> ```
> from cloudharness.utils.config import conf
> app_conf = conf.get_applications()
> for app in app_conf:
>     if app['deployment']['auto']:
>         # DO SOMETHING
> ```
> 
> about comments:
> can you please be a bit more specific about this?

I removed all outdated comments, all other comments are here above. if a thread is not resolved, it is still open.

libraries/cloudharness-common/cloudharness/utils/env.py
zsinnema: backwards compatibility

zsinnema:
> > > Thanks @zsinnema , a few questions/change requests:
> > > 
> > > * is bootstrap.sh.tar here on purpose?
> > > * the api requested in the card is not implemented
> > > * some previous comments have not been addressed/answered
> > 
> > 
> > bootstrap.sh.tar removed, on accident committed. may be we add an issue to remove the files after it has been run/used
> > api is almost the same, only some other naming
> 
> Almost the same is not same! We need some control especially in the common library, different people contribute here and different project can use it so the style must be kept consistent and issues should be a source of truth for the project history.
agree but in this case I think the developed api is more consistent because it's not about applications, it's about the configuration. therefore the package is called config instead of applications
> 
> > ```
> > from cloudharness.utils.config import conf
> > app_conf = conf.get_applications()
> > for app in app_conf:
> >     if app['deployment']['auto']:
> >         # DO SOMETHING
> > ```
> > 
> > 
> > about comments:
> > can you please be a bit more specific about this?
> 
> I removed all outdated comments, all other comments are here above. if a thread is not resolved, it is still open.

libraries/cloudharness-common/cloudharness/utils/config.py
zsinnema: done in my latest commit

zsinnema:
included in #36

PR #60, opened 2020-08-26
Feature/54

### LINKED ISSUE #55: Implement image overriding
In the case an application overrides another application, if no overriding docker image is defined, there is no way to override files added during the build. A common use case for this is changing the theme (e.g. a logo, a css, file), but it also would be useful to override code.

For instance take the following structure:

- applications
  - accounts
     - theme
- cloud-harness
  - applications
    - accounts
       - Dockerfile
       - theme
       - ...

What we can do now is to override the Dockerfile and do anything needed there, but that requires to copy the whole content.
- applications
  - accounts
     - Dockerfile
     - theme
     - ...
- cloud-harness
  - applications
    - accounts
       - Dockerfile
       - theme
       - ...

Another option is to base the image on the previously built image defining explicitly the dependency on the Dockerfile and then copying the theme, like:

```
ARG REGISTRY
ARG TAG=latest
FROM ${REGISTRY}accounts:${TAG}

COPY theme .
```

Desired outcome is that any file we change in the overriding folder is used in replacement of the original file.
A solution for this is to merge the directories in a temporary folder before actually performing the build, and use that folder as context for the image build.

This must be implemented consistently for the codefresh and the scripted build.

rodriguez-facundo:
Behaviour:

- Find all apps in the project application folder and in cloudharness application folder.
- Proceed to merge those folders that have the same name using the following rule:
  - cloudharness application folder is used as base
  - files in the other folder are copied to cloudharness folder and they override existing files if found
  - the rule above also applies to dockerfiles

When the script is used in a local machine, a tmp folder is used for the merge. When the script runs as part of CF pipeline, then the merge is done in place and the folder that is not the cloudharness one is deleted so that only one image is built and deploy downstream.

close #55


filippomc (approved):
Looks good!

PR #65, opened 2020-09-04
Feature/64

rodriguez-facundo:


filippomc:
@zsinnema fyk the new version of the gatekeeper (louketo)  supports certificated from a single command. So we don't need to setup anything custom anymore

filippomc (approved):
Looks good to me

zsinnema:
> @zsinnema fyk the new version of the gatekeeper (louketo) supports certificates with a single command. So we don't need to setup anything custom anymore

@rodriguez-facundo does this also replace the generation of the certificates via the .../cloudharness_utilities/scripts/bootstrap.sh ?

PR #79, opened 2020-10-08
#78 Add option to disable tls

### LINKED ISSUE #78: Add option to disable tls
Using https can be annoying when testing in Minikube, as dependent applications need to be opened and authorized in the browser before anything.

It would be useful to have the option to use plain http to be used for local testing and demos.

filippomc:
Close #78


zsinnema:
lgtm

PR #98, opened 2020-11-11
Feature/96 Support multiple paths in codefresh generate script

### LINKED ISSUE #96: Add multiple path support to codefresh.yaml generation
Also replace the hardcoded CLOUD_HARNESS generation

filippomc:
Closes #96

### DIFF
--- a/libraries/cloudharness-common/cloudharness/workflows/argo.py
+++ b/libraries/cloudharness-common/cloudharness/workflows/argo.py
@@ -19,7 +19,7 @@ plural = 'workflows'
 # determine the namespace of the current app and run the workflow in that namespace
 from cloudharness.utils.config import CloudharnessConfig as conf
 ch_conf = conf.get_configuration()
-namespace = ch_conf.get('namespace','argo-workflows')
+namespace = ch_conf and ch_conf.get('namespace','argo-workflows')
 
 CUSTOM_OBJECT_URL = f"/apis/{group}/{version}/{plural}"
 
@@ -152,7 +152,7 @@ try:
     check_namespace()
 except Exception as e:
     log.error('Namespace for argo workflows not found', exc_info=e)
-    log.info("Creating namespace " + namespace)
+    log.info("Creating namespace %s", namespace)
     try:
         create_namespace()
     except Exception as e:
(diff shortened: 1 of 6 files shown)


utilities/harness-codefresh
```diff
@@ -15,14 +15,18 @@ if __name__ == "__main__":
 
     parser = argparse.ArgumentParser(
         description='Walk filesystem inside ./applications and define build and deploy scripts.')
-    parser.add_argument('path', metavar='path', default=ROOT, type=str,
-                        help='Base path of the application.')
+    parser.add_argument('paths', metavar='paths', default=[ROOT], type=str, nargs='*',
+                        help='Base paths to start looking for applications. The paths will be processed '
+                             'in the given order so determining the build and overriding precedence.')
+    parser.add_argument('-i', '--include', dest='include', action="append", default=[],
```
zsinnema, line 21: very nice to have -i here too!

except one thing: do we need to add the base images to the -i too or will all base images always be build?
I btw. expect all base images always to be build ;-)

libraries/cloudharness-common/cloudharness/workflows/argo.py
```diff
@@ -152,7 +152,7 @@ def create_namespace():
     check_namespace()
 except Exception as e:
     log.error('Namespace for argo workflows not found', exc_info=e)
-    log.info("Creating namespace " + namespace)
+    log.info("Creating namespace %s", namespace)
```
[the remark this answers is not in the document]
filippomc: The %s is better for logging and this is how the loggin api is designed: the parameter is evaluated only if the logging level is adequate.

utilities/harness-codefresh
filippomc: this is at the moment behaving exactly as the harness-deployment counterpart, so if no -i is specified everything is built. Agree base images being included by default would be an improvement but I'd prefer to solve that with the application dependency definition rather than doing something now

libraries/cloudharness-common/cloudharness/workflows/argo.py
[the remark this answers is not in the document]
zsinnema: imho the f string evaluation is a nicer approach and they evaluate really performant. THe only advantage I see when the variable to evaluate (or function) is a really heavy computation.

utilities/harness-codefresh
zsinnema: okay

PR #108, opened 2020-11-23
make setup.py use requirements.txt for it's install_requires

### LINKED ISSUE #106: Update and pin 3rd party libraries of Cloud Harness common

zsinnema:
closes #106


libraries/cloudharness-common/setup.py
```diff
@@ -11,16 +11,14 @@
...
-]
-
+def _get_requirements(requirement_file):
+    with open(requirement_file) as f:
+        reqs = []
+        for l in f.read().splitlines():
+            reqs.append(l)
+        return reqs
 
+REQUIREMENTS = _get_requirements("requirements.txt")
```
filippomc, line 21: Although it may seem handy I prefer to avoid mapping setup requirements with requirements.txt. The reason is that setup.py is used in a broader env with possibly other libraries using the same dependencies and we want to avoid conflicts on the same library by using broader and safe-enough. Requirements.txt is instead used by the main application to define a very strict set of well-tested dependencies (usually generated with pip freeze > requirements.txt)

PR #116, opened 2020-12-04
#115 chore: add/update/delete kc user attributes

zsinnema:


libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py
```diff
@@ -252,3 +252,49 @@ def current_user_has_client_role(self, client_name, role):
...
+            attributes = user.get('attributes', [])
+            attributes[attribute_name] = attribute_value
+            admin_client.update_user(
+                user_id,
+                {
+                    'attributes': attributes
+                })
+            return True
+        except:
+            return False
```
filippomc, line 277: I think that it would be better to wrap and rethrow the exception. Using the return value to state success is so old C-like!

filippomc (changes requested):
all good, I only suggest to improve the exception handling

libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py
zsinnema: done

filippomc (approved):

PR #140, opened 2020-12-18
#109 codefresh dev and production deployment generation implemented

### LINKED ISSUE #109: Add codefresh tagging/pushing step and production deployment
Add to the a new parallel step "publish" to the [codefresh build](https://codefresh.io/docs/docs/codefresh-yaml/steps/push/) that pushes the images tagged as "latest" and with a custom tag specified as a variable in the build.

Also make the `harness-codefresh` script generate 2 files, one for build-development and one for production.
The production script can be parametrized with a tag variable and will deploy a previously built deployment.

Another related change: use the parent's application name in docker image repository so that different (potentially overridden) builds of cloudharness images won't conflict.

filippomc:
Closes #109 

After this pr, the codefresh pipeline must be changed to point to either 
- deployment/codefresh-dev.yaml
- deployment/codefresh-prod.yaml

Also check your application deployment-configuration: codefresh-template.yaml splits into 

- codefresh-template-dev.yaml
- codefresh-template-prod.yaml

The content must be adapted from the content inside codefresh-template-dev.yaml and the current codefresh-template.yaml

### DIFF
--- a/utilities/cloudharness_utilities/codefresh.py
+++ b/utilities/cloudharness_utilities/codefresh.py
@@ -2,9 +2,9 @@ import os
 import oyaml as yaml
 import logging
 
-from .constants import HERE, BUILD_STEP_BASE, BUILD_STEP_STATIC, BUILD_STEP_PARALLEL, BUILD_STEP_INSTALL, \
-    CODEFRESH_PATH, CODEFRESH_BUILD_PATH, \
-    CODEFRESH_TEMPLATE_PATH, APPS_PATH, STATIC_IMAGES_PATH, BASE_IMAGES_PATH, DEPLOYMENT_PATH
+from .constants import HERE, CF_BUILD_STEP_BASE, CF_BUILD_STEP_STATIC, CF_BUILD_STEP_PARALLEL, CF_STEP_PUBLISH, \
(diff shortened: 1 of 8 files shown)


utilities/cloudharness_utilities/codefresh.py
```diff
@@ -45,48 +55,71 @@ def codefresh_build_step_from_base_path(base_path, build_step, fixed_context=Non
...
 
-def codefresh_build_spec(**kwargs):
+def codefresh_template_spec(template_path, **kwargs):
     """
     Create Codefresh build specification
     :return:
     """
 
-    build = get_template(CODEFRESH_BUILD_PATH)
+    build = get_template(template_path)
```
zsinnema, line 93: I would do here ```return get_template(template_path).update(kwargs)```

filippomc: update is a mutating method and returns null.

filippomc: update is a mutating method and returns None

zsinnema: hahah yeah you're right (shame)

Issue #141, opened 2020-12-18
Install argo as part of the main deployment


filippomc:
At this moment, argo is installed separately in the namespace argo.
Changing it to install in the current namespace without other manual actions.

PR #147, opened 2021-01-05
Feature/145 fix new deployment

### LINKED ISSUE #145: Fix problems with then new generated codefresh files
* template overwrites not working
* timeout helm has no units
* duplicate step names because of publish step uses sames names as build
* publish step fails when running a codefresh build

filippomc:
Closes #145 
Also contains #141 argo deployment - still some problems to fix behind the gatekeeper


zsinnema (approved):
lgtm

Issue #149, opened 2021-01-14
Create support (proxy) for calling services on the current service url 


zsinnema:
we want to call other services from the current host so that the current frontend is unaware of where the other service resides

example if we want to call the config endpoint on the api of mnp-checkout `checkout.domain.xyz/api/config` but we are on host dashboard.domain.xyz we want to call the checkout endpoint on `dashboard.domain.xyz/proxy/mnp-checkout/api/config`

PR #151, opened 2021-01-14
#149 chore: implement extra paths for proxies to other services

zsinnema:


utilities/cloudharness_utilities/deployment-configuration/value-template.yaml
```diff
@@ -24,6 +24,8 @@ harness:
     auto: true
     name:
     port: 80
+  use_services:
+    # e.g. - name: mnp-checkout
```
filippomc, line 28: The empty value is invalid in yaml. Also better remove mnp reference. 

I can suggest for the empty array: 

`use_services: []`

filippomc (changes requested):
Looks good, requesting a small change request on the default yaml value.

filippomc (approved):

Issue #136, opened 2020-12-17
Production mode deployment configuration


filippomc:
`harness-deployment` production mode deployment will potentially set custom configuration and integrate the deployment with different setup, like replicas, backups, etc.

PR #163, opened 2021-01-26
#136 Base environment specific variables implementation

filippomc:
Now can run 
`harness-deployment ... -e [ENV]` and specific values-[ENV].yaml will be used to override some values in each application.

See implementation in samples applications, where values-prod.yaml is specified overriding some resource requests.


zsinnema (approved):
lgtm

PR #165, opened 2021-01-27
Feature/157 postgres and mongo backup

lrebscher:
* DB Backup is stored on a different persistence volume claim than DB
* Backup Deployment and Backup PVC only created if DB type is postgres
* Ensured that Pod runs as postgres user not as root

Manual Testing
* Tested new backup pod by specifying "@every 5m" as schedule
* Checked /backups folder if backup was created
* Verify created files: Copied DB backup to local filesystem, uploaded again to DB pod and imported backup as new DB name with `psql db_name < dump.sql`, worked without problem

### DIFF
--- a/utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
+++ b/utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
@@ -1,0 +1,36 @@
+/bin/bash << 'EOF'
+KEEP_DAYS=${BACKUP_KEEP_DAYS}
+KEEP_WEEKS=`expr $(((${BACKUP_KEEP_WEEKS} * 7) + 1))`
+KEEP_MONTHS=`expr $(((${BACKUP_KEEP_MONTHS} * 31) + 1))`
+
+mkdir -p "${BACKUP_DIR}/daily/" "${BACKUP_DIR}/weekly/" "${BACKUP_DIR}/monthly/"
+DFILE="${BACKUP_DIR}/daily/`date +%Y-%m-%d-%H%M%S`${BACKUP_SUFFIX}"
+WFILE="${BACKUP_DIR}/weekly/`date +%G-%V`${BACKUP_SUFFIX}"
+MFILE="${BACKUP_DIR}/monthly/`date +%Y-%m`${BACKUP_SUFFIX}"
+
+# Dump mongo database
+/usr/bin/mongodump -h $DB_HOST -u $DB_USER -p $DB_PASS --archive=$DFILE --gzip
+ls backups/*
+
+# Use hardlink instead of copy to save space
+if [ -d "${DFILE}" ]; then
+    WFILENEW="${WFILE}-new"
+    MFILENEW="${MFILE}-new"
+    rm -rf "${WFILENEW}" "${MFILENEW}"
+    mkdir "${WFILENEW}" "${MFILENEW}"
+    ln -f "${DFILE}/"* "${WFILENEW}/"
+    ln -f "${DFILE}/"* "${MFILENEW}/"
+    rm -rf "${WFILE}" "${MFILE}"
+    mv -v "${WFILENEW}" "${WFILE}"
+    mv -v "${MFILENEW}" "${MFILE}"
+else
+    ln -vf "${DFILE}" "${WFILE}"
+    ln -vf "${DFILE}" "${MFILE}"
+fi
+
+# Clean old files
+echo "Cleaning backups older than ${KEEP_DAYS} ..."
+find "${BACKUP_DIR}/daily" -maxdepth 1 -mtime +${KEEP_DAYS} -name "*${BACKUP_SUFFIX}" -print -delete
+find "${BACKUP_DIR}/weekly" -maxdepth 1 -mtime +${KEEP_WEEKS} -name "*${BACKUP_SUFFIX}" -print -delete
+find "${BACKUP_DIR}/monthly" -maxdepth 1 -mtime +${KEEP_MONTHS} -name "*${BACKUP_SUFFIX}" -print -delete
+EOF
--- a/utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
+++ b/utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
@@ -1,9 +1,20 @@
-/bin/bash << 'EOF'
+#!/bin/bash
+
+set -e
+set -u
+
 KEEP_DAYS=${BACKUP_KEEP_DAYS}
 KEEP_WEEKS=`expr $(((${BACKUP_KEEP_WEEKS} * 7) + 1))`
 KEEP_MONTHS=`expr $(((${BACKUP_KEEP_MONTHS} * 31) + 1))`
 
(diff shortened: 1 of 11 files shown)


lrebscher:
@filippomc  was thinking about adding an "active" flag to the backend yaml category.

```yaml
backup:
   active: false
```
 
Probably don't want to have backup pod running in a local setup. And with the new env configuration, a user could enable it only for prod by setting it in `values-prod.yaml` `active: true`.
 
 What do you think?

filippomc:
It would be really useful to add the active flag to the root of the values (rather than application specific)

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database.yaml
```diff
@@ -110,6 +110,121 @@ spec:
         persistentVolumeClaim:
           claimName: {{ .app.harness.database.name | quote }}
 ---
+{{ if eq .app.harness.database.type "postgres" }}
```
filippomc, line 113: Would be nice to have this split in different files: one file per database/backup:
like auto-database-postgres.yaml, auto-database-postgres-backup.yaml

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database.yaml
```diff
@@ -110,6 +110,121 @@ spec:
...
+            - name: POSTGRES_HOST
+              value: {{ .app.harness.database.name | quote }}
+            - name: POSTGRES_DB
+              value: {{ (index .app.harness.database .app.harness.database.type).initialdb | quote }}
+            - name: POSTGRES_USER
+              value: {{ .app.harness.database.user | quote }}
+            - name: POSTGRES_PASSWORD
+              value: {{ .app.harness.database.pass | quote }}
+            - name: SCHEDULE
+              value: {{ .app.harness.database.backup.schedule | quote }}
```
filippomc, line 146: Backup schedule would be better to keep as a global thing, so that all databases are in sync by design

utilities/cloudharness_utilities/deployment-configuration/value-template.yaml
```diff
@@ -71,3 +71,16 @@ harness:
       limits:
         memory: "2Gi"
         cpu: "1000m"
+    backup:
```
filippomc, line 74: Better move to global context (see above)

lrebscher: @filippomc  Do you want the resources definition as well to be globally configured? It might be useful to have it on the application level, since resources might be modified depending on specific DB backup tool and DB size.

filippomc: I don't see much on application level. Maybe something depending on the type of database we are doing the backup, but a common sizing that fits on every db would be probably enough

filippomc (approved):
Looks good to me, will wait also for @zsinnema to review it before merging

utilities/cloudharness_utilities/deployment-configuration/value-template.yaml
lrebscher: Okay, I will move it to the global level

utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
```diff
@@ -0,0 +1,35 @@
+/bin/bash << 'EOF'
```
zsinnema, line 1: may be add
```
#!/bin/bash

set -e
set -u
```

zsinnema: or do you like here documents more

utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
```diff
@@ -0,0 +1,35 @@
...
+    ln -f "${DFILE}/"* "${MFILENEW}/"
+    rm -rf "${WFILE}" "${MFILE}"
+    mv -v "${WFILENEW}" "${WFILE}"
+    mv -v "${MFILENEW}" "${MFILE}"
+else
+    ln -vf "${DFILE}" "${WFILE}"
+    ln -vf "${DFILE}" "${MFILE}"
+fi
+
+# Clean old files
```
zsinnema, line 30: I would do cleanup before create so we have more diskspace free before dumping

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database-mongo-backup.yaml
```diff
@@ -0,0 +1,55 @@
...
+  schedule: {{ .root.Values.backup.schedule | quote }}
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: "{{ .app.harness.database.name }}-backup"
+              image: {{ .app.harness.database.mongo.image }}
+              imagePullPolicy: IfNotPresent
+              command: ["/bin/sh"]
```
zsinnema, line 16: why not bash?

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database-mongo.yaml
```diff
@@ -0,0 +1,26 @@
+{{- define "deploy_utils.database.mongo" }}
+        image: {{ .app.harness.database.mongo.image }}
+        env:
+          - name: MONGO_INITDB_ROOT_USERNAME
+            value: {{ .app.harness.database.user | quote }}
+          - name: MONGO_INITDB_ROOT_PASSWORD
```
zsinnema, line 6: when we change the password we will have a problem
I know this is a general issue and we need to solve it in a generic way
I was thinking of:
* use of secrets
* use of configmap

what do you think and do you have other ideas?

zsinnema (changes requested):

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database-mongo-backup.yaml
lrebscher: no particular reason, switched to bash

utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
lrebscher: sounds good

utilities/cloudharness_utilities/deployment-configuration/helm/files/db-mongo-backup.sh
lrebscher: I used here documents to have multiline container args with proper interpretation of conditional statements, but I noticed it's not necessary anymore, since I extracted the script to a separate file and used `.Files.Get `

Your suggestion is nice, I applied it!

zsinnema (approved):

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database-mongo.yaml
lrebscher: I would go with secrets to store any kind of confidential data, it's as far as I know what K8s recommends and has already some nice functionality and support for various secret types, used it in previous projects and found it to be a good choice

PR #167, opened 2021-01-27
Added user roles to the single and all user(s) getters

ddelpiano:


zsinnema:
please also add these functions to make the implementation of the realm roles complete:
```
get_user_realm_roles
get_current_user_realm_roles
user_has_realm_role
current_user_has_realm_role
```

zsinnema (changes requested):
see my comments

libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py
```diff
@@ -330,6 +358,17 @@ def user_has_client_role(self, user_id, client_name, role):
         roles = [user_client_role for user_client_role in self.get_user_client_roles(user_id, client_name) if user_client_role['name'] == role]
         return roles != []
 
+    def user_has_realm_role(self, user_id, role):
+        """
+        Tests if the user has the given role within the given client
```
zsinnema, line 363: "within the current realm"

libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py
```diff
@@ -343,6 +382,17 @@ def current_user_has_client_role(self, client_name, role):
             client_name,
             role)
 
+    def current_user_has_realm_role(self, role):
+        """
+        Tests if the current user has the given role within the given client
```
zsinnema, line 387: "within the current realm"

zsinnema (changes requested):

ddelpiano:
@zsinnema I am merging this to fix the ifn deployment since currently we don't get clinician due to the logic to check the clinician or reviewer roles.
I fixed the latest changes requested already but if there is anything that you like me to change let me know and I will open a separate PR. Thx

Issue #155, opened 2021-01-18
Implement readiness and liveness probes


filippomc:
Implementation proposal:

1. Add a section on application's values.yaml specifying the probe http address:

```yaml
harness:
   readinessProbe:
        path: /api/ready
   livenessProbe:
        path: /api/live
```

2. Add the api paths and default implementation to harness-application generator.

The application developer will be able to implement the desired logic in readiness and liveness probes, keeping in mind the risks of using the probes to manage dependencies: https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/#kuberneteslivenessandreadinessprobes

PR #169, opened 2021-01-28
#155 Add liveness and readiness probe

lrebscher:
livenessProbe and readinessProbe are optional for the application user.
If either of them is specified, a httpGet probe is defined and the user must specify a path for it.
Optionally, she can also specify periodSeconds or initialDelaySeconds to further adjust the probes to the characteristic of the app or environment.

**Manual testing**
* Create application with `harness-application dbapp -t webapp -t db-postgres`
* Choose different configuration for liveness and readiness configuration & test if they are reflected in K8s deployment after helm upgrade


filippomc (approved):
Exactly what we needed, nice and clean, thanks!

Issue #168, opened 2021-01-28
update sentry implementation to be able to use a (one) sentry dsn instead of dsn's per app


zsinnema:

filippomc:
@zsinnema can you please write an article in the wiki on how to setup sentry? There is something to do also application side that is quite obscure to me at the moment. I'm afraid that we end up not using it if is not clear.

PR #174, opened 2021-02-01
#168 chore: implement external smtp server and remove postfix from repo also some code cleanup

zsinnema:
…po also some code cleanup


zsinnema:
the global dsn support can be used to set one dsn for the whole "solution" instead of a dsn per application of the solution
this is how you set it:
create a new folder in the solution applications folder named "common"
add a values.yaml to this folder with the values:
```
harness:
  env:
  - name: SENTRY_DSN
    value: <the dsn>
```

filippomc:
@zsinnema can you please write an article in the wiki on how to setup sentry? There is something to do also application side that is quite obscure to me at the moment. I'm afraid that we end up not using it if is not clear.

zsinnema:
> @zsinnema can you please write an article in the wiki on how to setup sentry? There is something to do also application side that is quite obscure to me at the moment. I'm afraid that we end up not using it if is not clear.

sure I can do, can you please detail what you mean by "quite obscure"?

filippomc:
> > @zsinnema can you please write an article in the wiki on how to setup sentry? There is something to do also application side that is quite obscure to me at the moment. I'm afraid that we end up not using it if is not clear.
> 
> sure I can do, can you please detail what you mean by "quite obscure"?

It's not clear to me what need to do to setup Sentry in the application in general and what are the concepts to keep in mind (dsn is one of them)

zsinnema:
> > > @zsinnema can you please write an article in the wiki on how to setup sentry? There is something to do also application side that is quite obscure to me at the moment. I'm afraid that we end up not using it if is not clear.
> > 
> > 
> > sure I can do, can you please detail what you mean by "quite obscure"?
> 
> It's not clear to me what need to do to setup Sentry in the application in general and in practice and what are the concepts to keep in mind (dsn is one of them)

I've created a new issue for that see https://github.com/MetaCell/cloud-harness/issues/175

PR #182, opened 2021-02-04
#181 Improvements to template loading and overriding

### LINKED ISSUE #181: Improve templating system
- [x] Override top level templates when --env is set on harness-deployment
- [x] Use helm/values.yaml as default template so as to allow simpler documentation

filippomc:
Now can define `deployment-configuration/values-template-[ENV].yaml` to set env-specific values.

Closes #181


zsinnema (approved):
lgtm

PR #185, opened 2021-02-09
Optional tagging on stage/production

filippomc:
With this change when DEPLOYMENT_PUBLISH_TAG is not defined on production pipeline we don't do the registry and git tagging


zsinnema (approved):
lgtm

Issue #186, opened 2021-02-09
Keycloak: Don't require ssl if -dtls is passed to harness-deployment


lrebscher:
The default realm configuration sets:
`"sslRequired": "external",`

[From the docs](https://wjw465150.gitbooks.io/keycloak-documentation/content/server_installation/topics/network/https.html):
> external requests
Keycloak can run out of the box without SSL so long as you stick to private IP addresses like localhost, 127.0.0.1, 10.0.x.x, 192.168.x.x, and 172..16.x.x. If you don’t have SSL/HTTPS configured on the server or you try to access Keycloak over HTTP from a non-private IP adress you will get an error.

If we want to allow non-private IP addresses, we have to set sslRequired to `none`.

Can be done with Helm template language.

PR #188, opened 2021-02-09
#186 Don't require TLS in Keycloak if TLS is deactivated

lrebscher:
**Manual testing with minikube:**
* Delete existing cluster
* Start new cluster
* Run `harness-deployment . -l -b -dtls` and install helm chart
* Check in Keycloak UI if ssl required is set to "none"

Repeat procedure but without "-dtls" flag and check if ssl required is set to "external requests".

**Important to know:**
When I enabled/disabled -dtls on an existing cluster, changes won't be reflected, since the realm.json is only imported at startup of accounts pod and only if it wasn't already imported. While one can manually delete keycloak realm and then restart the accounts pod, the tls option is used in many different templates and on first sight it seems that e.g. ingress controller must be restarted as well. **This is not done automatically after deploying the modified helm chart!**
For now, a fresh deployment is required so that a user can be sure that his changes to realm.json were applied.

### DIFF
--- a/applications/accounts/deploy/resources/realm.json
+++ b/applications/accounts/deploy/resources/realm.json
@@ -2,7 +2,7 @@
         "id": {{ .Values.namespace | quote }},
         "realm": {{ .Values.namespace | quote }},
         "enabled": true,
-        "sslRequired": "external",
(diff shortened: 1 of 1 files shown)


filippomc (approved):
Nice smart solution. Good for me

filippomc (approved):

PR #194, opened 2021-02-10
Feature/164 application naming

lrebscher:
**Proposal:** Application name must start and end with lowercased alphabetical character and may contain "-" as separator.

Would suggest to use a strict naming pattern. This way we avoid complexity and problems caused by different technologies requiring different patterns (K8s vs Python). The suggested pattern works good with K8s and for Python we  modify it in the background.

@filippomc @zsinnema  what do you think?


filippomc (approved):
Looks good and safe to merge

zsinnema (approved):
lgtm too

PR #204, opened 2021-02-15
#203 Add target port for auto-services

### LINKED ISSUE #203: Add targetPort to services
With no target port we are limited to use the same port for service and deployment.

filippomc:
Closes #203


zsinnema (approved):
lgtm

PR #209, opened 2021-02-16
Adding auth to SERVER_URL

ddelpiano:
Tested manually since this was rejecting the connection to my keycloak within minikube from the backend running on localhost.


zsinnema (approved):
lgtm

PR #210, opened 2021-02-16
Feature/152 cleanup workflow volumes

lrebscher:
@filippomc  see `operations.py` L70-74, the remaining changes were just PEP-8  formatting

I don't think we need anything else besides that to ensure cleanup of workflow volumes

Consider the values I chose as examples, I think it's useful to set `secondsAfterFailure` higher than the others so users have some time for investigation


zsinnema (approved):
lgtm

filippomc (approved):

PR #215, opened 2021-02-17
#196 Remove duplicated copy

### LINKED ISSUE #196: Folder copy in Application Template Backend Dockerfile
In the base Dockerfile for backend application (see [here](https://github.com/MetaCell/cloud-harness/blob/b14bc33622f30868c4e29509ae63cb6875acab49/utilities/cloudharness_utilities/application-templates/server/backend/Dockerfile#L7)), it copies the current folder, then runs pip install and copies again the current folder. 

First time should probably be copying only the requirements.txt, or am I missing sth?

lrebscher:
Closes #196

### DIFF
--- a/infrastructure/base-images/cloudharness-base/Dockerfile
+++ b/infrastructure/base-images/cloudharness-base/Dockerfile
@@ -12,7 +12,6 @@ RUN apk add gcc libc-dev g++ python3-dev libffi-dev openssl-dev rust musl-dev ca
 COPY libraries/cloudharness-common /libraries/cloudharness-common
 COPY libraries/client/cloudharness_cli /libraries/client/cloudharness_cli
 
-# RUN pip install -r /libraries/cloudharness-common/requirements.txt
 RUN pip install /libraries/cloudharness-common
 RUN pip install /libraries/client/cloudharness_cli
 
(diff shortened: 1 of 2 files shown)


infrastructure/base-images/cloudharness-base/Dockerfile
```diff
@@ -12,7 +12,6 @@ RUN apk add gcc libc-dev g++ python3-dev libffi-dev openssl-dev rust musl-dev ca
 COPY libraries/cloudharness-common /libraries/cloudharness-common
 COPY libraries/client/cloudharness_cli /libraries/client/cloudharness_cli
 
-# RUN pip install -r /libraries/cloudharness-common/requirements.txt
```
filippomc, line 15: Why this removed?

lrebscher: looked to me like commented code that isn't used anymore and replaced by the pip install statement below it, is it still needed?

filippomc: Yes, requirements.txt is different from the dependencies in setup.py because it's meant to keep stricter dependencies and also to track sub-dependencies. So the best practice in production is to install from requirements.txt before installing the package.
It's also useful to first copy/install requirements.txt before copy/install the rest to make a better use of the docker caching

filippomc (approved):

PR #238, opened 2021-02-23
Feature/234

### LINKED ISSUE #234: Improve backup volume management
- Use same volume for all backups
- mount volume on db pods for inspection - manual recovery

filippomc:
Closes #234 

Backup volume is now unique for all applications and the volume is mounted to database pods for easier inspection and restore.
Also default backup time changed to 5 minutes.


libraries/cloudharness-common/tests/test_infrastructure.py
```diff
@@ -2,6 +2,8 @@
 set_test_environment()
 
 from cloudharness.infrastructure import k8s
-
+kubectl_enabled = False
```
lrebscher, line 5: This violates PEP8, there must be a newline between variable module level function

utilities/tests/resources/applications/myapp/deploy/values.yaml
```diff
@@ -0,0 +1,7 @@
+harness:
+  name: "I'm useless"
```
lrebscher, line 2: Eh okay 😆 I feel sorry for that app

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database-postgres-backup.yaml
```diff
@@ -38,9 +38,9 @@ spec:
                 - name: BACKUP_SUFFIX
                   value: {{ .root.Values.backup.suffix | quote }}
                 - name: BACKUP_DIR
-                  value: {{ .root.Values.backup.dir | quote }}
+                  value: {{ (printf "%s/postgres/%s" .root.Values.backup.dir .app.harness.database.name) | quote }}
```
lrebscher, line 41: This could backfire since one can specify the same database name in two applications causing accidental overwrite of backups, without even noticing it until one needs that backup.

If backups should be stored on one volume across dbs, we need to ensure that one DB backup cannot conflict with another DB backup.

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database.yaml
```diff
@@ -46,27 +60,24 @@ spec:
         volumeMounts:
           - name: {{ .app.harness.database.name | quote }}
             mountPath: /data/db
+          {{- if .root.Values.backup.active }}
+          - name: "db-backup"
+            mountPath: /backup
```
lrebscher, line 65: hmm, this way a single DB container would have access to backups of all DBs, I think we should limit this somehow

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database-postgres-backup.yaml
filippomc: usually applications shouldn't specify the database name as it is automatically assigned from app name. But we really need it to be configurable for backwards compatibility (otherwise we cannot use auto-database with application that use a db with different name than default)

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database.yaml
filippomc: I'm not a privacy/security expert but I don's see backups to be visible to other applications as a problem. I think it's just handy to have all backups in the same place to look at all system backup or copy all backups to a different place.

libraries/cloudharness-common/tests/test_infrastructure.py
filippomc: Did this and other formatting fixes

utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-database.yaml
zsinnema: agree with @filippomc , security wise, if a "hacker" has access to the cluster he/she can access all the volumes already

zsinnema (approved):
lgtm

PR #239, opened 2021-02-24
#226 chore: accounts set max heap size

### LINKED ISSUE #226: Robust resource configuration for Keycloak (accounts)
If we want to adjust the memory settings of keycloak, we have to pass -Xms and -Xmx to keycloak.

Keycloak is a JVM application and will set per default `-Xms64m -Xmx512m`. 

1. Even if we specify 1024m as resource limit, it will never use it.
2. If we specify anything below 512m it crashes, since it doesn't care about the K8s resource configuration (see #224 )

for JVM Applications, we can configure heapsize by using the `JAVA_OPTS` environment variable.
See this article for an example: https://medium.com/@ravindrashekhawat5876/keycloak-heap-size-management-on-kubernetes-8f504ccf40cd

=========================================================================
  JBoss Bootstrap Environment
  JBOSS_HOME: /opt/jboss/keycloak
  JAVA: java
  JAVA_OPTS:  -server -Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true  --add-exports=java.base/sun.nio.ch=ALL-UNNAMED --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED --add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED

zsinnema:
closes #226

### DIFF
--- a/applications/accounts/deploy/values.yaml
+++ b/applications/accounts/deploy/values.yaml
@@ -34,6 +34,8 @@ harness:
       value: "user"
     - name: DB_PASSWORD
       value: "password"
+    - name: JAVA_OPTS
+      value: -server -Xms64m -Xmx1024m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true  --add-exports=java.base/sun.nio.ch=ALL-UNNAMED --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED --add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED
   database:
     auto: true
     name: keycloak-postgres
--- a/applications/accounts/deploy/values.yaml
+++ b/applications/accounts/deploy/values.yaml
@@ -35,7 +35,7 @@ harness:
     - name: DB_PASSWORD
       value: "password"
     - name: JAVA_OPTS
-      value: -server -Xms64m -Xmx1024m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true  --add-exports=java.base/sun.nio.ch=ALL-UNNAMED --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED --add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED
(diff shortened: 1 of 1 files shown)


zsinnema:
used the java opts from the current keycloak pod and modified the max heap size `-Xmx`
```
09:51:39,179 INFO  [org.jboss.as] (MSC service thread 1-1) WFLYSRV0050: Keycloak 11.0.2 (WildFly Core 12.0.3.Final) stopped in 116ms
=========================================================================

  JBoss Bootstrap Environment

  JBOSS_HOME: /opt/jboss/keycloak

  JAVA: java

  JAVA_OPTS:  -server -Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true  --add-exports=java.base/sun.nio.ch=ALL-UNNAMED 
--add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED --add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED

=========================================================================
```

applications/accounts/deploy/values.yaml
```diff
@@ -34,6 +34,8 @@ harness:
       value: "user"
     - name: DB_PASSWORD
       value: "password"
+    - name: JAVA_OPTS
+      value: -server -Xms64m -Xmx1024m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djboss.modules.system.pkgs=org.jboss.byteman -Djava.awt.headless=true  --add-exports=java.base/sun.nio.ch=ALL-UNNAMED --add-exports=jdk.unsupported/sun.misc=ALL-UNNAMED --add-exports=jdk.unsupported/sun.reflect=ALL-UNNAMED
```
lrebscher, line 38: would set the maximal heap size 10-20% lower than memory set in limit request to reserve some space for other processes

applications/accounts/deploy/values.yaml
```diff
@@ -34,6 +34,8 @@ harness:
       value: "user"
     - name: DB_PASSWORD
       value: "password"
+    - name: JAVA_OPTS
```
lrebscher, line 37: in a more recent keycloak image version than we currently, a `JAVA_OPTS_APPEND` env variable was introduced. They introduced this because there is some risk of values being omitted if JAVA_OPTS is defined directly: https://github.com/keycloak/keycloak-containers/pull/282/files

Cannot say if this is a problem here, but might be a good idea to upgrade and use `JAVA_OPTS_APPEND`

zsinnema: since we don't use the newer versions this doesn't seem a problem to me

applications/accounts/deploy/values.yaml
zsinnema: fine for me, see new commit

lrebscher (approved):
lgtm

PR #243, opened 2021-02-25
#242 adding userGroups and realmRoles to members

ddelpiano:


filippomc (approved):
lgtm

zsinnema (commented):
@ddelpiano what is the reason to add these to this call?
with many many many users this call will become very slow
please explain how the results of this call will be used in the application

ddelpiano:
> @ddelpiano what is the reason to add these to this call?
> with many many many users this call will become very slow
> please explain how the results of this call will be used in the application

@zsinnema we are using this to get all the users that belongs to a certain organization/clinic, for consistency IFN is expecting the user returned to have the same format.

zsinnema (approved):

PR #250, opened 2021-03-09
#249 pod affinity added to argo workflows

### LINKED ISSUE #249: Add pod affinity to argo workflows
When using the same volume already in use by another pod in write mode with ReadWriteOnce the pod addinity can be used to ensure that the pods are in the same node.

To do so, can add to the spec something like the following:

```yaml
spec
  affinity:
    podAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchExpressions:
          - key: security
            operator: In
            values:
            - S1
        topologyKey: topology.kubernetes.io/zone
```

filippomc:
Closes #249


zsinnema (approved):
lgtm

PR #273, opened 2021-05-06
Fix build when no registry is set

filippomc:


zsinnema (approved):
tested and works 
lgtm

Issue #240, opened 2021-02-25
Handle secrets


filippomc:
Helm overrides secrets values, so secrets cannot be included in the deployment unless we provide secrets values directly in the deployment itself.

Possible solutions:

- Include secrets to the deployment as parameters. These parameters may have a default value and can be set as variables in the codefresh deployment
- Use helm lookup functionality to include the secret if does not exist.

filippomc:
Example lookup:

```yaml
{{- if not (lookup "v1" "Secret" .Values.namespace "my-secret") }}
apiVersion: v1
kind: Secret
metadata:
  name: my-secret
  namespace: {{ .Values.namespace }}
type: Opaque
data:
  # Mail credentials need to be set manually by the sys admin
  a: Q0hBTkdFTUU=
  b: Q0hBTkdFTUU=
  c: Q0hBTkdFTUU=
{{- end }}
```

filippomc:
Integration with Codefresh variables is important: we use the values coming from codefresh to set secrets

filippomc:
Harness deployment allows to specify the secrets. Using a .env file would allow to specify secret inside the pipeline and also on local deployment

zsinnema:
if we let helm manage the secrets then we always need to run a deployment in case we want to change the value of the secret.
not sure if this is the best way to do

e.g. if a MNP Checkout stripe token is changed (because of reasons) we need to run the whole pipeline to get the secret where it resides in to get deployed. this sounds strange and not logical to me

question: why not have "default" secrets created via helm (as above shown) and don't let an automated process touch them anymore

filippomc:
The way I'd interpret deployment set secret is as default values. With the lookup we can ensure that secrets don't get overridden after the first deploy

filippomc:
After some tests I think this can be a good strategy:

- Secrets are managed automatically by cloudharness, as deployments, services, etc. Secret values come from values.yaml
- The deployment pipeline sets default values for all secrets using variables, with something like:

```yaml
  deployment:
    stage: deploy
    image: codefresh/cfstep-helm:3.4.1
    title: Installing chart
    environment:
      - CHART_REF=./deployment/helm
      - RELEASE_NAME=${{NAMESPACE}}
      - KUBE_CONTEXT=${{CLUSTER_NAME}}
      - NAMESPACE=${{NAMESPACE}}
      - CHART_VERSION=${{DEPLOYMENT_PUBLISH_TAG}}
[...]

PR #284, opened 2021-05-28
#240 chore: implemented secrets through mounting instead of env vars and make sure secret values aren't in the allvalues.yaml

zsinnema:
update the secrets usage from:
```
get_secret("mnp-checkout", "stripe-public-key")
```
to
```
get_secret("stripe-public-key")
```

also no need to do a b64decode anymore
e.g.
```
base64.b64decode(get_secret("mnp-checkout", "stripe-public-key")).decode("ascii")
```
needs to be replaced by
```
get_secret("stripe-public-key")
```

### DIFF
--- a/libraries/cloudharness-common/cloudharness/utils/secrets.py
+++ b/libraries/cloudharness-common/cloudharness/utils/secrets.py
@@ -1,5 +1,13 @@
 import os
 
+SECRETS_PATH = os.getenv("CH_SECRETS_PATH", "/opt/cloudharness/resources/secrets")
+
+
+class SecretNotFound(Exception):
+    def __init__(self, secret_name):
+        Exception.__init__(self, f"Secret {secret_name} not found.")
+
+
 def get_secret(name: str) -> str:
     """
     Helper class for the CloudHarness application secrets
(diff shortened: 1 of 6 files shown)


libraries/cloudharness-common/cloudharness/utils/secrets.py
```diff
@@ -10,4 +10,10 @@ def get_secret(name: str, key: str):
...
         key (str): name of the data key in the secret
     """
-    return os.environ.get(f'{name}-{key}', None)
+    try:
+        secrets_path = '/opt/cloudharness/resources/secrets'
+        with open(os.path.join(secrets_path, name)) as fh:
+            return fh.read()
+    except:
+        # if no secrets folder or file exists
+        return ''
```
zsinnema, line 19: @filippomc or do we here want to raise an exception?

filippomc: yes definitely this is a runtime exception

filippomc (changes requested):
Integration required: automatically generate helm variable replacements in codefresh pipelines.

Should be something like:


```yaml
  deployment:
    stage: deploy
    image: codefresh/cfstep-helm:3.4.1
    title: Installing chart
    environment:
    ....
    - CUSTOM_apps_[APPNAME]_secrets_github_token=${{GITHUB_TOKEN}}
    - CUSTOM_apps_[APPNAME]_secrets_github_user=${{GITHUB_USER}}
```

filippomc (approved):
Looks good to me

PR #289, opened 2021-06-08
Feature/287: Fix image dependencies using Skaffold

### LINKED ISSUE #287: Build dependencies not working with Skaffold
Skaffold provides an artifact dependencies mechanism that allows to define dependencies between image buildings.
In Cloudharness we can define build dependencies in values.yaml and this is reflected in the Skaffold configuration but the default value [is taking over the Skaffold value](https://github.com/GoogleContainerTools/skaffold/issues/5943) so images are not being updated as expected.

A solution can be to remove the default values from all dockerfiles, like:

```
ARG REGISTRY
ARG TAG=latest
ARG CLOUDHARNESS_FLASK=${REGISTRY}cloudharness-flask:${TAG}
FROM $CLOUDHARNESS_FLASK
```

```
ARG CLOUDHARNESS_FLASK
FROM $CLOUDHARNESS_FLASK
```

but that will require some work not to break the legacy and the codefresh build pipelines that rely on REGISTRY and ARG at the moment.

filippomc:
Closes #287 

After this pr, image dependencies can be specified directly like:

```
ARG CLOUDHARNESS_BASE
FROM $CLOUDHARNESS_BASE as base
```

Legacy specification (REGISTRY and TAG arguments) still works


zsinnema (approved):
lgtm

Issue #297, opened 2021-06-24
EventClient add create topic in produce in case topic doesn't exist


zsinnema:
It could happen that the topic is not (yet) created and the client wants to send (produce) messages
in that case a KafkaTimeoutError is raised

PR #298, opened 2021-06-24
#297 chore: add create topic to produce when kafkatimeout error occurs & refrased some error messages

zsinnema:
…s & refrased some error messages


filippomc (approved):
lgtm

PR #311, opened 2021-07-07
#310 expandable deployment pipelines implemented

### LINKED ISSUE #310: Improve multiple deployment pipeline script generation
There is logic at the moment that assumes that we only have a "dev" and a "prod" deployment in place.

Proposing here to allow any deployment script to be generated, leaving dev and prod as default ones.

The pipeline files will be generated only when the -e flag is used for that specific pipeline, e.g

`harness-deployment ... -e prod` will generate `codefresh-prod.yaml` file, using all `*-prod.yaml` file overrides where defined.

filippomc:
Closes #310


zsinnema (approved):
nice improvement

lgtm

PR #341, opened 2021-10-05
fix: update collection related imports

sanjayankur31:
Moved to collections.abc in py3.3+

Reference: https://docs.python.org/3/library/collections.abc.html


filippomc (approved):
Looks good, thanks Ankur!

PR #351, opened 2021-10-25
Fix databases versions

### LINKED ISSUE #350: Fix Postgres default version
The postgres default image is now postgres:latest. This causes issues when new postgres upgrades.

For instance, upgrading from 13 to 14 raises the error:

> data directory was initialized by PostgreSQL version 13, which is not compatible with this version 14.0

The 13 is the safest option for backwards compatibility

filippomc:
Fixes the problem with pod restarts are at risk on major database update versions, as images are all using the tag latest.

Closes #350

### DIFF
--- a/utilities/cloudharness_utilities/deployment-configuration/value-template.yaml
+++ b/utilities/cloudharness_utilities/deployment-configuration/value-template.yaml
@@ -65,7 +65,7 @@ harness:
     pass: metacell
     # -- settings for mongo database (for type==mongo)
     mongo:
-      image: mongo:latest
+      image: mongo:5
       ports:
         - name: http
           port: 27017
@@ -78,7 +78,7 @@ harness:
           port: 5432
     # -- settings for neo4j database (for type==neo4j)
     neo4j:
-      image: neo4j:latest
+      image: neo4j:4.1.9
       memory:
         size: 256M
         pagecache:


libraries/cloudharness-common/cloudharness/utils/server.py
```diff
@@ -110,6 +110,7 @@ def handle_exception(e: Exception):
             except:
                 logging.error("Error checking sentry configuration", exc_info=True)
                 data['trace'] = traceback.format_exc()
+            logging.error(str(e), exc_info=True)
```
zsinnema, line 113: I don't see why this is needed here. If we want to directly log the issue in Sentry then please use the Sentry sdk to create a Sentry message (f.e. use `capture_message` from the sentry_sdk)

zsinnema (changes requested):

utilities/cloudharness_utilities/deployment-configuration/value-template.yaml
```diff
@@ -65,20 +65,20 @@ harness:
     pass: metacell
     # -- settings for mongo database (for type==mongo)
     mongo:
-      image: mongo:latest
+      image: mongo:5
```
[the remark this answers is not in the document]
filippomc: applications can override deployment-configuration values at global level. Every application can override this as wanted, what is wrong with latest tag is that if the value is not overridden in the application, major upgrades can break the database deployment by accident. I think that having a fixed default is preferable

libraries/cloudharness-common/cloudharness/utils/server.py
filippomc: I don't think Sentry is a replacement of base server logging: Sentry is an optional and nice to have addition.

filippomc:
@zsinnema until we have a better idea than using latest tag to make version updates I think we should go with this. I think it's too dangerous to risk major database updates but just restarting a pod

filippomc:
Notice: if applications are already on upper versions of the database will need to tweak the version before upgrading

zsinnema (approved):

PR #357, opened 2021-12-17
#356 ssh:default support for docker buildkit builds to support cloning of private git repos

zsinnema:
For buildkit set `ssh:default` to allow private repo to be cloned in a docker file
see https://docs.docker.com/develop/develop-images/build_enhancements/#using-ssh-to-access-private-data-in-builds for more info


filippomc (approved):
lgtm

Issue #362, opened 2021-12-29
Security vulnerability, upgrade KC to >=15.1.1


zsinnema:
see https://www.keycloak.org/2021/12/cve

Issue #361, opened 2021-12-28
Use a “role” to connect to keycloak for the CH library instead of the admin/metacell password


zsinnema:
I want to change the password of the admin. Now the password is in the git repo and everybody who has access (public) can read it. When changing the admin password the pods using the CH library need also be updated

zsinnema:
instead of using a "role" we could also opt for a 2nd username/password pair stored in a secret and generated using the initial username/password. This second pair will only be used for the API. The username/password pair could be auto generated on initial application deployment (as a secret). It will not be stored in the git repo

filippomc:
@zsinnema added the alert as this breaks already existing deployment: the user is not created automatically if it does not exist.

PR #363, opened 2021-12-29
#362 chore: upgrade Keycloak to 16.1.0

zsinnema:
I've tested the new CH version with MNP and everything seems to work as it should.

### DIFF
--- a/applications/accounts/Dockerfile
+++ b/applications/accounts/Dockerfile
@@ -1,3 +1,3 @@
-FROM quay.io/keycloak/keycloak:11.0.2
+FROM quay.io/keycloak/keycloak:16.1.0
 # Customize keycloak look
 COPY themes/custom /opt/jboss/keycloak/themes/custom


filippomc:
The upgrade didn't work for me on a hot replacement (see error below), while it worked after deleting the previous deployment.
This is an issue for the business continuity and may create a lot of waste time to the unaware developer. 
We should think about a way to properly track update issues and in general major version updates in the release notes.
Also the issue #362 does not really explain the reason of this major update.

I want to take this kind of potential upgrade issues seriously, because if we start having uncontrolled issues the response will be not to upgrade the cloudharness version at all.

```
10:12:18,245 ERROR [org.infinispan.CLUSTER] (thread-6,ejb,accounts-78dc8c5fdd-m2bfs) ISPN000474: Error processing request 1@accounts-bcd77d457-ph7h6: org.infinispan.commons.CacheException: Unknown command id 90!
at org.infinispan@10.1.8.Final//org.infinispan.commands.RemoteCommandsFactory.fromStream(RemoteCommandsFactory.java:214)
at org.infinispan@10.1.8.Final//org.infinispan.marshall.exts.ReplicableCommandExternalizer.readCommandHeader(ReplicableCommandExternalizer.java:99)
...
```

zsinnema:
I ran the upgrade on IFN staging and no error was shown and keycloak was updated to v16

here you can find the issue that this pr is solving: https://www.keycloak.org/2021/12/cve

filippomc:
> I ran the upgrade on IFN staging and no error was shown and keycloak was updated to v16
> 
> here you can find the issue that this pr is solving: https://www.keycloak.org/2021/12/cve

Thanks for the specification. What are your thoughts about tracking potential issues like this? I'm thinking about adding a tag to the issues so it's easier to track them in the release notes

zsinnema:
@filippomc do you mean how to generate release notes? In GitHub it's possible to create a release and there you can generate the release notes by getting all commit messages (same you can do with plain git). What do you think of adding the url to the git commit message?

filippomc:
@zsinnema I've added some tags to github about the scope and general warnings. I think we need something more high level in the release notes than commit messages to make it clear at a glance if there is something to worry about when upgrading the ch version.

filippomc:
#361 also upgrade keycloak. Should we closing.

PR #364, opened 2021-12-30
Feature/361: stop using the admin account for api calls to KC

zsinnema:


applications/accounts/scripts/create_api_user.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ "${api_user_secret_exists}" -ne "0" ];
+then
+set -e
+echo Creating API user
+
+# create the user and reload keycloak
+/opt/jboss/keycloak/bin/add-user-keycloak.sh -u ${USERNAME} -p ${PASSWORD}
+
+kubectl create secret generic -n ${NAMESPACE} accounts-api \
```
filippomc, line 17: not sure we want to allow this in the pod

filippomc (changes requested):
I like the solution of using a secret to access a separate admin user.
What I don't fully understand is why the secret is created at runtime, with the assumption that the pod has permission to create (and potentially modify?) secrets, which is a potential security flaw to avoid by setting a lower permissions to the pods.
What about adding the secret as part of the deployment instead like we do for other secrets?

applications/accounts/scripts/create_api_user.sh
zsinnema: changed it, see commit below

applications/accounts/deploy/templates/api-user-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
filippomc, line 1: This is fine but why not use the harness/secret? Shouldn't it be the default way to create secrets?

filippomc (approved):
Tested on osb: good to go for me

filippomc:
Notice: this pr updates keycloak, so may have issues on hot deployments

applications/accounts/deploy/templates/api-user-secret.yaml
zsinnema: @filippomc I think we can't set a random value with the default CH secrets , only a fixed string. One of our goals is to not store passwords in git so a fixed string isn't the best.

filippomc: note: probably we should allow to create random passwords in the default CH secrets

zsinnema: @filippomc yeah I was thinking that too :-)

filippomc: https://github.com/MetaCell/cloud-harness/issues/387

zsinnema: implemented random secret values and use CH secret managment here: https://github.com/MetaCell/cloud-harness/pull/389

Issue #384, opened 2022-01-21
Accounts secret should be mounted only on deployments requiring it


filippomc:
The accounts secret containing the credentials to access keycloak apis should be mounted only where it is required, so applications having a hard dependency from accounts.
The current behaviour is not ideal because shares potentially sensible information where it's not explicitly required, also breaks the deployments where the accounts application is not included as part of the deployment.

filippomc:
Reopened as there is a logged error: when only by importing cloudharness, it looks for the accounts secret, even if the accounts client is not needed. The secret should be read only the moment it is required.

```
W
2022-02-08 18:25:04,849 [ERROR] __init__.<module>: Error on cloudharness configuration. Check that the values file /opt/cloudharness/resources/allvalues.yaml your deployment.
Traceback (most recent call last):
  File "/libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py", line 23, in get_secret
    with open(os.path.join(AUTH_SECRET_PATH, name)) as fh:
FileNotFoundError: [Errno 2] No such file or directory: '/opt/cloudharness/resources/auth/api_user_password'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py", line 40, in <module>
    PASSWD = get_secret("api_user_password")
  File "/libraries/cloudharness-common/cloudharness/auth/keycloak/__init__.py", line 27, in get_secret
    raise AuthSecretNotFound(name)
cloudharness.auth.keycloak.AuthSecretNotFound: Secret api_user_password not found.
```

PR #385, opened 2022-01-21
Feature/216-384

### LINKED ISSUE #216: Wrong image name is used if application is based solely on base template
**Reproduction**

* Run `harness-application my-app`  -> creates an application that uses only the base template.
* Run `harness-deployment -l -b -n ch -i my-app` -> docker builds image `my-app-backend`. But `values.yaml` sets `my-app` as image name
* Pod won't be able to pull image since it's expecting `my-app` image to be present

filippomc:
Closes #216 and #384

Not that the fix for #384 may break applications that use the keycloak apis but do not include accounts as a hard dependencies.


utilities/cloudharness_utilities/deployment-configuration/helm/templates/auto-deployments.yaml
```diff
@@ -105,9 +105,11 @@ spec:
           - name: cloudharness-allvalues
             mountPath: /opt/cloudharness/resources
             readOnly: true
+          {{- if has  "accounts" .app.harness.dependencies.hard }}
```
zsinnema, line 108: @filippomc did you add accounts as a hard dependency to the apps that need the secret? for backwards compatibility, otherwise this is a breaking change

zsinnema: and also to mnp

filippomc: Yes this can break applications do not declare the dependency, that's why you see the tag alert:deployment. At the same time, an application that uses the keycloak admin api in its logic has always been intended to declare the hard dependency to accounts. Now by chance other applications could include account and everything works, but that's just why the odds are good this is true that things usually work regardless the undeclared dependency.
That's probably just a semantic detail, but I'd rather use the breaking-change tag for real changes in the api or more in general the way things are intended to be.

zsinnema: agree

zsinnema (approved):

zsinnema (approved):

PR #389, opened 2022-01-25
#361 chore: use CH secrets for api acocunt secret, use a random string for nullable secrets default value

### LINKED ISSUE #387: Secrets auto create random passwords
Now secrets can be added to applications values files like:

```
harness:
   secrets:
      my-secret-value: "value"
```

This declares the secrets but requires the value to be set statically or in the deployment pipeline.

For the cases where the secret is just internal it would be better to just leave the value empty and let the system create it for us.

Use the strategy used on accounts application: https://github.com/MetaCell/cloud-harness/blob/a71b59e462477ba727904aebf84451372e6e75dc/applications/accounts/deploy/templates/api-user-secret.yaml#L8. Also change accounts to use the new secret declaration strategy

zsinnema:
Closes #387

### DIFF
--- a/applications/accounts/deploy/templates/api-user-secret.yaml
+++ b/applications/accounts/deploy/templates/api-user-secret.yaml
@@ -1,8 +1,0 @@
-apiVersion: v1
-kind: Secret
-type: Opaque
-metadata:
-  name: accounts-api
-data:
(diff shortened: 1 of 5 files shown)


filippomc (approved):
Looks good to me

PR #391, opened 2022-01-25
#390 chore: reset the k8s resource values

zsinnema:


filippomc (changes requested):
I can imagine why we want to reduce the resources for the accounts application but need to be properly documented in the issue. Please also add scope tags to the issue and link issue-pr.

zsinnema:
> I can imagine why we want to reduce the resources for the accounts application but need to be properly documented in the issue. Please also add scope tags to the issue and link issue-pr.

@filippomc it's resetting it back to it's original values. This commit (https://github.com/MetaCell/cloud-harness/commit/24866ea2e2cf6fea43279a5ba1257e660cbc4c40) messed it up

filippomc:
> @filippomc it's resetting it back to it's original values. This commit ([24866ea](https://github.com/MetaCell/cloud-harness/commit/24866ea2e2cf6fea43279a5ba1257e660cbc4c40)) messed it up

Yes I understand what it's for but let's write that in the issue and make proper linking and assignment so that we keep the issue history clear

zsinnema:
> > @filippomc it's resetting it back to it's original values. This commit ([24866ea](https://github.com/MetaCell/cloud-harness/commit/24866ea2e2cf6fea43279a5ba1257e660cbc4c40)) messed it up
> 
> Yes I understand what it's for but let's write that in the issue and make proper linking and assignment so that we keep the issue history clear

I've linked the related issues and the commit this commit reverts too.

Issue #397, opened 2022-01-27
Secret accounts is not created because api_user_password is empty


zsinnema:
currently the api_user_password is set to `api_user_password:`
this will cause the accounts secrets not bein created by helm

for Helm to operate correctly it should be set to `api_user_password: ""`

PR #398, opened 2022-01-27
#397 chore: accounts secrets not created

zsinnema:


filippomc (approved):
approving to solve the actual issue, with the hope that we can get the empty version working soon

Issue #412, opened 2022-02-01
Deployment error: failed to mount cloudharness_kc_accounts secret


zsinnema:
There is a deployment error in case an application doesn't use Keycloak (ch accounts service):
```
MountVolume.SetUp failed for volume "cloudharness-kc-accounts" : secret "accounts" not found
```

This is because of the fact that the `cloudharness-kc-accounts` secret is always mounted.

Required fix:
only mount the `cloudharness-kc-accounts` secret if there is a "hard" dependecy to `accounts` configured in the values.yaml

filippomc:
Duplicate of #384

PR #413, opened 2022-02-01
#412 fixes deployment error in case app doesn't depend on accounts

zsinnema:


zsinnema:
already fixed here: https://github.com/MetaCell/cloud-harness/pull/385

PR #470, opened 2022-03-22
#469 feature: enable application volume to be shared across multiple different deployments

### LINKED ISSUE #469: Enable mounting of a shared PVC into multiple deployments
For some applications it would be really nice if we could mount a "shared" PVC into multiple deployments.
There are apps where the frontend and the backend are different deployments. E.g. the frontend is a NXING serving a static react app and the backend is a Flask/Django server.

For user content (uploaded content) the application would benefit from sharing a PVC between the frontend and the backend.
The backend will store the uploaded content (e.g. file) and the frontend will serve the file back to the user.

We can do this by adding a `volume` object to the harness.deployment of the microservices. The issue is that when I have 2  or more services the PVC is multiple times created.

solution proposal:
add the attribure `auto` to the harness.deployment.volume object.
when true: create the PVC
when false: don't create the PVC

zsinnema:
Closes #469 

Implemented solution: add extra attribute to the deployment.volume object: auto true/false

How to test this PR: create 2 microservices and add the same volume to both with the exception that only one should have auto set to true

### Sanity checks:
- [x] From the issue and the current PR it is explicitly stated how to test the current change
- [x] All the linked issues are included in one milestone

### DIFF
--- a/docs/applications/README.md
+++ b/docs/applications/README.md
@@ -95,6 +95,7 @@ The most important configuration entries are the following:
   - `deployment`: creates a deployment
     - `auto`: if true, creates the deployment automatically
     - `resources`: define cpu and memory limits
+    - `volume`: application persistent volume
   - `service`:
     - `auto`: if true, creates the service automatically
   - `dependencies`: lists of applications/images this application depends from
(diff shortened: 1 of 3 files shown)


filippomc (approved):
Nice PR, thanks!

PR #488, opened 2022-05-13
Fix samples application

filippomc:
This is fixing a regression from some previous commit in the samples application, which was not working anymore


zsinnema (approved):
lgtm

PR #633, opened 2022-11-22
#632 feat: add switch to retrieve group attributes

### LINKED ISSUE #632: Add possibility to retrieve the Keycloak attributes of the groups the user is member of
It would be nice if the CH Common auth keycloak package could return the group attributes of a user to. This will make it possible to use Keycloak to store some "generic" / "group level" settings that can be used for micro services to act on

zsinnema:
Closes #632 

Implemented solution: added param `brief_representation` (default True) to the get user / get group , to retrieve the group details (incl. the attributes) set the param to False 

How to test this PR: 
```python
from cloudharness.auth.keycloak import AuthClient
ac=AuthClient()
ac.get_users(brief_representation=False)
```

### Sanity checks:
- [x ] From the issue and the current PR it is explicitly stated how to test the current change
- [x] All the linked issues are included in one milestone

- [x] The changes included in this pr are out of the current test coverage scope

### DIFF
--- a/libraries/cloudharness-common/cloudharness/auth/keycloak.py
+++ b/libraries/cloudharness-common/cloudharness/auth/keycloak.py
@@ -250,9 +250,14 @@ class AuthClient():
         return True
 
     @with_refreshtoken
-    def get_group(self, group_id, with_members=False) -> UserGroup:
+    def get_group(self, group_id, with_members=False, brief_representation=False) -> UserGroup:
         """
         Return the group in the application realm
+        
+        :param group_id: the group id to get
+        :param with_members: Default False, when set to True all members of the group are also retrieved
+        :param brief_representation: Default False, when set to True all attributes of the group are also retrieved
+
 
         GroupRepresentation
         https://www.keycloak.org/docs-api/16.0/rest-api/index.html#_grouprepresentation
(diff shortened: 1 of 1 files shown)


libraries/cloudharness-common/cloudharness/auth/keycloak.py
```diff
@@ -250,9 +250,14 @@ def create_client_role(self, client_id, role):
         return True
 
     @with_refreshtoken
-    def get_group(self, group_id, with_members=False) -> UserGroup:
+    def get_group(self, group_id, with_members=False, brief_representation=True) -> UserGroup:
```
filippomc, line 253: The api looks a little bit inconsistent as there is already the `with_members` attribute. It would be more natural a parameter like `with_attributes` with default `false`

zsinnema: it's not only attributes :-), like realroles and client roles.... we could do `with_details` what do you think?

filippomc: looks good

zsinnema: done

filippomc (approved):

PR #718, opened 2024-01-25
Add tag / build reference api

filippomc:
Closes [CH-47](https://metacell.atlassian.net/browse/CH-47)

Implemented solution: the build id (as git commit hash) is now extracted by harness-deployment and added to values.yaml.
The common service exposes it and the tag.
Added an example on samples microservice
image

How to test this PR: start samples.DOMAIN and verify that is shows the tag and build id as in the screenshot.

### Sanity checks:
- [ ] From the issue and the current PR it is explicitly stated how to test the current change
- [ ] All the linked issues are included in one milestone
- [ ] All the linked issues are in the Review/QA column of the [board](https://app.zenhub.com/workspaces/cloud-harness-5fdb203b7e195b0015a273d7/board)

[CH-47]: https://metacell.atlassian.net/browse/CH-47?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zsinnema (commented):
question: @filippomc isn't it the case that for CH all apps in the deployment have the same version? Then imo we only need to implement this only in one application / endpoint (f.e. common). Instead of in all apps 

or may be in the keycload login template then also apps like jupyter hub are covered

filippomc:
@zsinnema the version of this api is representative of the deployment so it's the same for all application. It is indeed implemented only for the common app, what the samples app does here is to display the value in the ui.

zsinnema:
> @zsinnema the version of this api is representative of the deployment so it's the same for all applications. It is indeed implemented only for the common app, what the samples app does here is to display the value in the ui.

what I meant @filippomc is that we don't need an application to show the version number. We can just show it in the keycloak login screen.....

applications/samples/frontend/src/components/Version.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { useState, useEffect } from 'react';
+
+
+
+const Version = () => {
+  const [result, setResult] = useState(null);
+  useEffect(() => {
```
zsinnema, line 7: @filippomc please move this to the keycloak login page template 
imo we should show the version there
and also we could show it on the keycloack account details page (the one of the user, where you can go to for f.e. changing your password/details)

zsinnema (changes requested):

filippomc:
@zsinnema I think it can be a good idea to show the version in the keycloak login screen, but we are not customizing the keycloak theme on ch, which makes updates simpler. Also derived applications have their theme and might not want to show the version there.

zsinnema (approved):

PR #720, opened 2024-01-31
Allow to clone an external repo before the build

filippomc:
Closes [CH-37](https://metacell.atlassian.net/browse/CH-37)

Implemented solution:

the clone is done before the build, hence on skaffold and codefresh

The [before-build hook](https://skaffold.dev/docs/lifecycle-hooks/#before-build-and-after-build) is used on Skaffold  

Additional clone steps are added on codefresh

How to test this PR: Add application git dependencies on values.yaml like
```yaml
harness:
  dependencies:
    git:
    - url: https://github.com/MetaCell/nwb-explorer.git
      branch_tag: master
      path: nwbexplorer
```
to one of your applications (say myapp).
Then run `harness-deployment ... -i myapp -e dev` and verify that:
- skaffold.yaml has a new hook with a clone command
- deployment/codefresh-dev.yaml is updated with a clone step for https://github.com/MetaCell/nwb-explorer.git

### Sanity checks:
- [x] From the issue and the current PR it is explicitly stated how to test the current change
- [x] All the linked issues are included in one Jira sprint and release
- [x] All the linked issues are in the Review/QA column state -- see [backlog](https://metacell.atlassian.net/jira/software/c/projects/CH/boards/33/backlog)

[CH-37]: https://metacell.atlassian.net/browse/CH-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/clone.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+REPOSRC=$2
+LOCALREPO=$3
+BRANCH=$1
+
+# We do it this way so that we can abstract if from just git later on
```
zsinnema, line 7: @filippomc not sure if its a "wise" idea to share the same git repo between different "apps"

f.e. nwbexplorer and netpyne both need repo "abc"
but netpyne needs branch "xyz"
and nwbe needs branch "qwe"

then there can be a situation where both builds run at the same time and then it's uncertain which branch is used by what build

my suggestion would be to clone the repo in a "app" specific folder instead of in a global repo folder

what do you think?

tools/deployment-cli-tools/ch_cli_tools/codefresh.py
```diff
@@ -33,6 +34,27 @@ def literal_presenter(dumper, data):
...
+        if "gitlab" in url:
+            return "gitlab"
+        if "bitbucket" in url:
+            return "bitbucket"
+        return "github"
+    except:
+        return "${{ DEFAULT_REPO }}"
+
+def clone_step_spec(conf: GitDependencyConfig, context_path: str):
+    return {
```
zsinnema, line 49: @filippomc question: can we also implement support for cloning private git repos?
I know for git we can do it using an access token but I prefer that we don't store the access token in the "deployment" git repo

tools/deployment-cli-tools/tests/resources/applications/myapp/deploy/values.yaml
```diff
@@ -6,6 +6,12 @@ harness:
       - legacy
     build:
       - cloudharness-flask
+    git:
+    - url: https://github.com/a/b.git
+      branch_tag: master
+    - url: https://github.com/c/d.git
+      branch_tag: v1.0.0
+      path: myrepo
```
zsinnema, line 14: @filippomc question: the path is this the path that will be copied/mounted in the Docker build?

tools/clone.sh
filippomc: @zsinnema agree and that's exactly what's happening, might not be clear from the code.
For example, this is how it looks like in codefresh:

```yaml
steps:
  main_clone:
    title: Clone main repository
    type: git-clone
    stage: prepare
    repo: '${{CF_REPO_OWNER}}/${{CF_REPO_NAME}}'
    revision: '${{CF_BRANCH}}'
    git: github
  post_main_clone:
    title: Post main clone
    type: parallel
    stage: prepare
    steps:
    - title: Cloning cloud-harness repository...
      type: git-clone
      stage: prepare
      repo: https://github.com/MetaCell/cloud-harness.git
      revision: '${{CLOUDHARNESS_BRANCH}}'
      working_directory: .
      git: github
    - title: Cloning nwb-explorer.git repository...
      type: git-clone
      repo: https://github.com/MetaCell/nwb-explorer.git
      revision: master
      working_directory: /home/user/OSBv2/cloud-harness/./applications/dependencies/nwb-explorer.git
      git: github
```

And in Skaffold:
```yaml
  - context: /home/user/OSBv2/cloud-harness/./applications/samples
    docker:
      buildArgs: {NOCACHE: '1706717709.5154254', REGISTRY: '', TAG: v0.2}
      dockerfile: applications/samples/Dockerfile
      ssh: default
    hooks:
      before:
      - command: [sh, tools/clone.sh, master, 'https://github.com/MetaCell/nwb-explorer.git',
          /home/user/OSBv2/cloud-harness/./applications/samples/dependencies/nwb-explorer.git]
```

tools/deployment-cli-tools/tests/resources/applications/myapp/deploy/values.yaml
filippomc: No, I've finally opted for not making assumptions on where the files are getting copied inside the Dockerfile, there are too many variables to take into account as copying the whole repo is not usually a good idea. 

That's the path where the repo is cloned to inside `dependencies`. It's optional as we use the repo name by default.

Please let me know if it's clear from the docs https://github.com/MetaCell/cloud-harness/blob/feature/CH-37/docs/applications/dependencies.md#git-repository-dependencies

tools/deployment-cli-tools/ch_cli_tools/codefresh.py
filippomc: Private repos should be supported in codefresh as it uses your login, While the local machine should work with git:// paths assuming that the ssh keys are set. Token based authentication looks potentially useful but my preference would be to put that support on a separate issue, unless there is already a practical use case

tools/clone.sh
zsinnema: @filippomc okay cool I understand.

Then my question changes: do we really want to "cache" the git repo in the app folder? or do we clean afterwards/before and do always a `git export`

the advantage is that the build can not change the branch by executing a "git checkout <new branch>"

so a thought

filippomc: I like the idea to keep the directory as I expect development to be ongoing there too: if the codebase were settled I'd rather go with a static dockerfile extension rather than this dynamic git clone

zsinnema: @filippomc I think I was not clear enough.

check this case:
* we use the script to clone a repo for example branch `develop`
* then in the Dockerfile we cd into the folder of this git repo
* then in the Dockerfile we do a `git checkout main`
resulting in a different branch than it was set in the dependencies

imo this should be prevented
easy solution is to `git export` instead of `git clone`

tools/deployment-cli-tools/tests/resources/applications/myapp/deploy/values.yaml
zsinnema: question @filippomc : what is the usecase for having a path here? It can only cause errors if I have 2 different repos using the same path (when I make a mistake)

filippomc: Yes it's only about avoiding remote and potential name clashes. Could have left it out but was fast to implement :)

zsinnema (approved):
lgtm

PR #735, opened 2024-02-23
CH-100 Add preliminary docker-compose support

aranega:
Closes CH-100

Adds a preliminary docker-compose support for cloud-harness.

Implemented solution: 
The implementation relies on helm templates dedicated for docker-compose.
A new folder can be defined and used for all docker-compose related templates, while a mechanism allows one to define where some files can be generated.
On a general note, all files that needs to be generated from templates and needs to be embedded in containers needs to be generated in a specific directory in the file system.
The generated files are then mounted in the countainer.

How to test this PR: 
There is a new option `--docker-compose` that can be passed as argument of the command line for the `harness-deployment` command.
This new option will generate in `deployment/compose` all the templates and specific files for the cloudharness deployment, a `deployment/docker-compose.yaml` will be produced, and the `skaffold.yaml` file is modified to target docker-compose for deployment (even if this feature is still not yet supported).

### Sanity checks:
- [X] The pull request is explicitly linked to the relevant issue(s)
- [ X The issue is well described: clearly states the problem and the general proposed solution(s)
- [X] From the issue and the current PR it is explicitly stated how to test the current change
- [X] The labels in the issue set the scope and the type of issue (bug, feature, etc.)
- [X] All the automated test checks are passing
- [ ] All the linked issues are included in one milestone
- [X] All the linked issues are assigned

- [X] The present changes do not change the preexisting api in any way

- [X] There is no reason why deployments based on CloudHarness may break after the current update


applications/nfsserver/deploy/templates-compose/nfsserver-deployment.yaml
```diff
@@ -0,0 +1,19 @@
+{{- define "nfsserver.deployment" }}
```
filippomc, line 1: not supporting this, removing

applications/events/deploy/templates-compose/deployments.yaml
```diff
@@ -0,0 +1,97 @@
+{{- define "events.deployment" }}
```
filippomc, line 1: if it doesn't work, better remove it

deployment-configuration/compose/templates/allvalues-template.yaml
```diff
@@ -0,0 +1,20 @@
+{{- /*
```
filippomc, line 1: Not sure what's the purpose of this template. I'm seeing two files generated -- allvalues.yaml and  allvalue2.yaml, with the same content as values.yaml

deployment-configuration/compose/templates/auto-compose.yaml
```diff
@@ -0,0 +1,241 @@
...
+      - "/var/run/docker.sock:/var/run/docker.sock:ro"
+      - "./certs/:/certs/:ro"
+      - "./traefik.yaml:/etc/traefik/dynamic_conf/conf.yml:ro"
+
+{{- range $app_name, $app_config := .Values.apps }}
+  {{- if has $app_name (list "argo" "events" "nfsserver") -}}
+  {{- continue -}}
+  {{- end -}}
+  {{ $deployment :=  $app_config.harness.deployment }}
+  {{- if eq $app_name "nfsserver" }}
```
filippomc, line 30: better remove, there's no use case of the nfs server for docker-compose (its purpose is to provide a ReadWriteMany volume on K8s, but all volumes are equivalent of ReadWriteMany on docker as we are on a single node)

deployment-configuration/compose/templates/auto-compose.yaml
```diff
@@ -0,0 +1,241 @@
...
+  {{- with $app_config.harness.deployment.volume }}
+  {{ .name }}:
+  {{- end }}
+  {{- with $app_config.harness.database }}
+  {{ .name }}:
+  {{- if eq .type "postgres" }}
+  dshm-{{ .name }}:
+  {{- end }}
+  {{- end }}
+  {{- if eq $app_name "nfsserver" }}
```
filippomc, line 234: remove

deployment-configuration/compose/templates/auto-gatekeepers.yaml
```diff
@@ -0,0 +1,146 @@
...
+    - {{ .app.harness.service.name }}:{{ .app.harness.service.name }}.default
+{{- end }}
+
+{{- define "securedservice.deploy.resources" }}
+{{- $tls := not (not .root.Values.tls) }}
+cloudharness-metadata:
+  path: resources/generated/{{ .app.harness.service.name }}-gk/proxy.yml
+
+data: |-
+    verbose: {{ .root.Values.debug }}
```
filippomc, line 49: So much duplication here, we should find a way to use the same source template for this

deployment-configuration/compose/templates/auto-gatekeepers.yaml
```diff
@@ -0,0 +1,146 @@
...
+  path: resources/generated/{{ .app.harness.service.name }}-gk/cacert.crt
+
+data: |-
+{{ .files.Get "resources/certs/cacert.crt" | indent 2 }}
+---
+cloudharness-metadata:
+  path: resources/generated/{{ .app.harness.service.name }}-gk/access-denied.html.tmpl
+
+data: |-
+    <!DOCTYPE html>
```
filippomc, line 91: Another duplicated file, better extract and share

deployment-configuration/compose/values.yaml
```diff
@@ -0,0 +1,79 @@
+# -- If set to true, local DNS mapping is added to pods.
```
filippomc, line 1: this is the same values.yaml we use for kubernetes, better avoid duplication

tools/deployment-cli-tools/ch_cli_tools/dockercompose.py
```diff
@@ -0,0 +1,891 @@
...
+from . import HERE, CH_ROOT
+from cloudharness_utils.constants import TEST_IMAGES_PATH, VALUES_MANUAL_PATH, HELM_CHART_PATH, APPS_PATH, HELM_PATH, \
+    DEPLOYMENT_CONFIGURATION_PATH, BASE_IMAGES_PATH, STATIC_IMAGES_PATH, COMPOSE
+from .utils import get_cluster_ip, get_image_name, env_variable, get_sub_paths, guess_build_dependencies_from_dockerfile, image_name_from_dockerfile_path, \
+    get_template, merge_configuration_directories, merge_to_yaml_file, dict_merge, app_name_from_path, \
+    find_dockerfiles_paths, find_file_paths
+
+from .models import HarnessMainConfig
+
+KEY_HARNESS = 'harness'
```
filippomc, line 27: These constants are also defined in helm.py. Better extract and share

tools/deployment-cli-tools/ch_cli_tools/dockercompose.py
```diff
@@ -0,0 +1,891 @@
...
+        logging.info(f'openssl container logs: {logs}')
+
+        # stop the container
+        container.kill()
+
+        logging.info("Created certificates for local deployment")
+
+    def __finish_helm_values(self, values):
+        """
+        Sets default overridden values
```
filippomc, line 399: this logic is the same as in helm.py, extracting a superclass might allow us to avoid duplication

filippomc (changes requested):
This is looking great functionally wise and far beyond the initial expectations. Before pulling it I'd like to ensure unit tests are there and we have done all our best to avoid code duplication.

deployment-configuration/compose/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
aranega: I see your point, but trying to share the files between the helm configuration files and the docker compose one will only make things more obscure to understand. 

Those two files (this one and the `values.yaml`) that are in the `deployment-configuration` needs to be duplicated in a way. If not, it means that there is a weird behavior that we should implement saying that the content of those files that are required for the template generation to work, need to be took from the `helm` folder for some reasons. It also means that we cannot rely on the current mechanism which copies files for a target from the corresponding folder, but instead, we would have a mix between "I pick files from the dedicated `compose` folder, but sometimes no, I need to pick them in the `helm` folder", which makes the generator more complicated and the logic of the configuration file generation more obscure. 

A lot of files appears to be duplicated, but they are here to support the simplicity of "when I generate a target, the files from the corresponding target folder are copied" instead of having a mix of "sometimes its took from this folder, sometimes no", which would also require to implement a mechanism to check if the files that needs to be picked from the `helm` folder needs to be copied only if it doesn't exists in the `compose` folder to support configuration overriding (or other target folder if we support others in the future). 

Another solution would be to share those files in a `common` or `shared-configuration` folder which would require to implement the same logic described above, but with another folder. I would vote against it as it makes the whole logic more complicated to understand, makes the generator more complex than just "copy files from the corresponding folder" and breaks, in a way, the isolation and independence of each target, to avoid having two files that looks the same. Why make it complicated when we can keep it simple and stupid?

deployment-configuration/compose/templates/auto-gatekeepers.yaml
aranega: In this kind of context, and same for the other files embedded in this same file, we could have a shared template folder somewhere where the content of the files that needs to be generate for k8s and docker-compose are added, and then included inside of the respective templates (for helm and docker-compose)? 

The way of declaring the content of a file is not the same between the k8s target and docker-compose target, so the only thing that we can "share" there is the content of the file itself, not the way it will be generated (I'm not sure I'm very clear in what I'm saying)

deployment-configuration/compose/templates/allvalues-template.yaml
aranega: It's a forgotten experiment that is here. I saw one template for the k8s target that was copying/reproducing the `allvalues.yaml` files, but I wasn't sure if it was replicating it exactly as the original file or if it was modifying entries at the same time. I understood later that the logic for this template is actually to have the content of the `allvalues.yaml` inside of the k8s cluster, while the approach of docker compose is different as we can mount the file in the container. I need to remove this generation, thanks for spotting it!

deployment-configuration/compose/Chart.yaml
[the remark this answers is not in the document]
filippomc: I see your point the reason to avoid duplication. I'm removing the irrelevant configuration from compose's values.yaml

tools/deployment-cli-tools/ch_cli_tools/dockercompose.py
```diff
@@ -0,0 +1,320 @@
...
+            #     'name': self.image_tag(img_name, build_context_path=task_path, dependencies=values[KEY_TASK_IMAGES].keys()),
+            #     # 'context': os.path.relpath(task_path, self.dest_deployment_path.parent),
+            #     # 'dockerfile': 'Dockerfile',
+            # }
+
+            values[KEY_TASK_IMAGES][task_name] = self.image_tag(img_name, build_context_path=task_path, dependencies=values[KEY_TASK_IMAGES].keys())
+
+        return values
+
+    def inject_entry_points_commands(self, helm_values, image_path, app_path):
```
filippomc, line 288: I'm not sure this is necessary: we do command overriding on skaffold to enable debugging that doesn't work well with unicorn

filippomc (approved):

PR #782, opened 2024-11-01
Feature/ch 157

condar-metacell:
Closes [CH-157](https://metacell.atlassian.net/browse/CH-157)

Added support for a django-ninja template when using harness-application and harness-generate. Also refactored the application build process into a more structured process of builder components.

... 

To test try generating applications using `harness-application -t django-ninja` and also using `harness-generate servers` for that app to see the generates servers and openapi schema.

...

# Sanity checks:


application-templates/django-ninja/README.md
```diff
@@ -0,0 +1,86 @@
...
+This application is composed of a Django-Ninja backend and a React frontend.
+
+### Backend
+
+Backend code is inside the *backend* directory.
+See [backend/README.md#Develop]
+
+### Frontend
+
+Backend code is inside the *frontend* directory.
```
D-GopalKrishna, line 33: DOC: `Frontend code`?

condar-metacell: Yep, should be, one moment...

condar-metacell: Fixed, same issue was present in the existing django template, so that's been updated as well.

D-GopalKrishna: thanks @condar-metacell !

application-templates/django-base/backend/django_baseapp/asgi.py
afonsobspinto, line 1: Same typo here:
`init_listner` -> `init_listener`

afonsobspinto (changes requested):

application-templates/django-ninja/Dockerfile
```diff
@@ -0,0 +1,35 @@
+ARG MNP_UI
```
afonsobspinto, line 1: Maybe there's a reason for the renaming but without any context on that my preference would be to keep the more generic name: CLOUDHARNESS_FRONTEND_BUILD instead

application-templates/django-base/backend/django_baseapp/asgi.py
afonsobspinto, line 1: I believe 
>os.environ.setdefault("DJANGO_SETTINGS_MODULE", "__APP_NAME__.settings")
should actually be:
>os.environ.setdefault("DJANGO_SETTINGS_MODULE", "django_baseapp.settings")
since that's the place where the settings are defined

application-templates/django-ninja/backend/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pydantic==2.9.2
+django-ninja
```
afonsobspinto, line 2: Would it make sense to have cloudharness and cloudharness_django as dependencies here? Since they are conceptually required to run the application (f.e. they are used in settings.py) @ddelpiano

ddelpiano: I would expect so. I understand we might need CH repo locally if I am in need of regen the template files or the codefresh pipeline, but if I am purely working on the backend I might not need CH and such dependency should be part of the requirements.txt so that all I need to run the backend is already covered by that. Thanks!

ddelpiano (changes requested):
requirements.txt is missing ch as dependency since the settings.py of the template requires it.

application-templates/django-ninja/backend/requirements.txt
filippomc: I agree it's better to be explicit about all the dependencies that are imported, even if they are inherited by the base image (like this case)

application-templates/django-ninja/Dockerfile
condar-metacell: Nah, this was a copy/paste from where we first implemented it in MNP, you're correct that the more generic name should be used.

application-templates/django-ninja/backend/requirements.txt
condar-metacell: I don't know how I'd go about adding that to the `requirements.txt` file. The cloud harness modules aren't on pip, so I can't just have `cloudharness-django` in there, and the path to them is not well defined; they're at `/libraries/cloudharness-django` in the base image, but that isn't necessarily going to always be true, and definitely won't be true for running it outside the image.

filippomc:
> requirements.txt is missing ch as dependency since the settings.py of the template requires it.

we are not adding inside dependencies on requirements.txt as they don't add anything to the install process there

PR #792, opened 2025-01-20
Release/2.5.0

filippomc:
Creates [v2.5.0](https://metacell.atlassian.net/projects/CH/versions/10241/tab/release-report-all-issues)


tools/deployment-cli-tools/ch_cli_tools/openapi.py
```diff
@@ -79,18 +129,61 @@ def generate_python_client(module, openapi_file, client_src_path, lib_name=LIB_N
...
+            yaml.dump(data, yaml_file)
+
+
+def generate_openapi_from_ninja_schema(app_name: str, app_path: pathlib.Path) -> None:
+    # check if cloudharness_django python library is installed
+    python_module = to_python_module(app_name)
+    try:
+        import cloudharness_django
+        # dynamicall import python_module
+        __import__(f'{python_module}')
```
condar-metacell, line 161: **Suggestion**

There is no need to wrap this in an f-string, i.e. `__import__(python_module)`

tools/deployment-cli-tools/ch_cli_tools/openapi.py
```diff
@@ -109,3 +202,153 @@ def get_dependencies():
...
+        client_lib_name: str,
+        client_types: ClientType,
+) -> None:
+    """
+    Generates client stubs
+    """
+    if not should_generate('client libraries'):
+        return
+
+    logging.info('Generating client libraries for %s', str(client_types))
```
condar-metacell, line 265: **Suggestion**

There is an inconsistent style to added logging lines. I think that because we're not dealing with structured logging or high performance code we should just use f-strings for readability, i.e. this line would be `logging.info('Generating client libraries for {client_types}')`.

tools/deployment-cli-tools/ch_cli_tools/utils.py
```diff
@@ -174,17 +176,23 @@ def replaceindir(root_src_dir, source, replace):
...
-            src_file = join(src_dir, file_)
+            src_file = pathlib.Path(src_dir) / file_
             replace_in_file(src_file, source, replace)
 
 
-def replace_in_file(src_file, source, replace):
-    if src_file.endswith('.py') or basename(src_file) == 'Dockerfile':
-        replace = to_python_module(replace)
-    with fileinput.FileInput(src_file, inplace=True) as file:
+def confirm(question, default_true=True):
```
condar-metacell, line 183: **Suggestion - Naming**

I don't think `default_true` is the best name for the parameter here. I think `default` would be better in this situation, or perhaps `default_on_non_answer` as that is specifically what it is used for.

condar-metacell, line 183: **Suggestion**

I think we should have type hints on here, so this would be `def confirm(question: str, default: bool = True) -> bool:`

tools/deployment-cli-tools/ch_cli_tools/utils.py
```diff
@@ -174,17 +176,23 @@ def replaceindir(root_src_dir, source, replace):
...
+    answer = input(f"{question} (Y/n): ").casefold()
+    return (default_true and not answer) or answer == "y"
```
condar-metacell, line 185: **Suggestion**

Rather than have this all on one line (that I find harder to parse) I'd split this up across multiple lines, e.g.

```python
if not answer:
    return default

return answer == "y"
```

or if you do want it on one line I think the following is better: 
```python
return answer == "y" if answer else default
```

Overall I just find the multiple clauses with parenthesis and negatives make this harder to parse than it needs to be, so the suggestions fix those.

tools/deployment-cli-tools/harness-generate
```diff
@@ -1,176 +1,127 @@
...
+    args = get_command_line_arguments()
     get_dependencies()
-    if args.models and os.path.exists(os.path.join(root_path, "libraries/models")) and (not args.interactive or input("Do you want to generate the main model? [Y/n]").upper() != 'N'):
-        generate_model()
-    if not (args.clients or args.models) or args.server:
-        generate_servers(root_path, interactive=args.interactive, server=args.server)
-    if not (args.server or args.models) or args.clients:
-        generate_clients(root_path, args.client_name, interactive=args.interactive)
+
+    if args.path:
```
condar-metacell, line 20: **Suggestion**

This should not be needed, the path argument has a default of `pathlib.Path.cwd()` configured, so the `else` of this should never be reached.

tools/deployment-cli-tools/harness-generate
```diff
@@ -1,176 +1,127 @@
...
+        root_path = args.path.absolute()
+
+    else:
+        root_path = pathlib.Path.cwd()
+
+       # Check if applications folder exists, if not, go up until it's found
+    while not (root_path / 'applications').exists():
```
condar-metacell, line 27: **Suggestion**

I don't think I agree with this walking up the directory tree to find an applications directory as it could lead to some really odd edge cases if the tool is used not as expected. I'd much rather have the script check for an applications directory, and if it doesn't exist simply exit with an error message, it will be easy enough to the user to either navigate to the right directory or create the applications directory before running again.

tools/deployment-cli-tools/ch_cli_tools/openapi.py
filippomc: This is the intended and most performant way to use the logging API with parameters. I don't mind is another style is used if performance is not a concern but a Python developer should get used to this.

tools/deployment-cli-tools/ch_cli_tools/utils.py
filippomc: Type hints are a great addition to recent Python but duck typing is still at the fundaments of the language, so I think we should not enforce other languages styles or enforce typing when obviously not needed just for the sake of consistency

tools/deployment-cli-tools/harness-generate
filippomc: This covers two important cases: 
- the application might be defined inside cloud harness or in a project outside it, on multiple levels
- a pretty common use case when the code is generated from inside the application directory

As I find the auto discovery pretty handy in common use cases and the edge error cases a quite weird misuse of the tool, I think it's better to leave it as it is

tools/deployment-cli-tools/ch_cli_tools/openapi.py
filippomc: btw it's not the first time I get a remark for not using f-strings in logs :D

tools/deployment-cli-tools/ch_cli_tools/utils.py
filippomc: This is an interesting argument :) I think it's really a matter of taste and habit. In Python, and/or logic is pretty common and one-liners are generally considered clear and elegant, although I agree your version is more readable on average. Wonder what our Pythonic champion @aranega thinks about this

aranega: There is generally no special receipe for dealing with `input`, indeed long boolean lines are usually a little bit more complicated to understand, but even if the `if ... return ...` would fit the PEP-8, I feel something is a little bit strange in general about the function. Perhaps the what's not so obvious is the fact that if you set the default to `False`, then the message would still be `Y/n`, emphasazing that "Y" is the default, and if I enter "Y", I would get false. 

Another way to do it would be to change the signature of `confirm` to have something like that (but not necessarily that):

```python
def confirm(question, yes="y", no="n", no_as_default=False):
    if no_as_default:
        yes, no = no, yes
    answer = input(f"{question} ({yes.capitalize()}/{no}) ").casefold() or yes
    return answer == yes
```

In a normal use this would be:

```python
>>> confirm("OK?")
OK? (Y/n)

>>> confirm("OK?", no_as_default=True)
OK? (N/y) 

>>> confirm("OK?", yes="yes", no_as_default=True)
OK? (N/yes) 

>>> confirm("OK?", yes="yes")
OK? (Yes/n)
```

filippomc: Thanks @aranega good point about matching the default with the message!

filippomc: Actually I'm feeling like the default parameter is creating more troubles than it should have, and haven't seen around a lot of personalization needs in confirmation messages, usually it's Y/n and that's it, and it's definitely all we need here.

Leaning towards

```python
def confirm(question):
    answer = input(f"{question} (Y/n): ").casefold()
    return answer == "y" if answer else True
```

aranega: That's a good point. There is only few command line tools that I use that give you the choice to have `(N/y)`, but the default is useful if you want to localize the app later (which is probably not the case here), to have, e.g: `(O/n)` for french.

For the code, I would go with that:

```python
def confirm(question):
    answer = input(f"{question} (Y/n): ").casefold() or "y"
    return answer == "y"
```
Or a 1 liner:

```python
def confirm(question):
    return (input(f"{question} (Y/n): ").casefold() or "y") == "y"
```
Or if really you want to keep the ternary `if`, you can do a 1 liner with the walrus operator

```python
def confirm(question):
    return answer == "y" if (answer := input(f"{question} (Y/n): ").casefold()) else True
```
But honestly, it's less easy to read :D (I don't like the walrus operator)

tools/deployment-cli-tools/ch_cli_tools/utils.py
condar-metacell: I honestly really disagree with this. Type hinting helps the development process by telling you the expected parameters. Sure this signature is simple and easy to read, but `default` will be a `bool`, it'll return a `bool`, the only thing you could handle for duck typing is maybe making `question` an `Any`.

I don't understand the reticence to embrace type hinting, it's not a new feature it's a decade old now, and I don't see any downside to adding hinting to a method like this, but can definitely see upsides. Anyway, not gonna hold up this PR for this comment as I've said my piece.

tools/deployment-cli-tools/ch_cli_tools/utils.py
condar-metacell: I'd say the following option @aranega provided is neat and readable:
```python
def confirm(question):
    answer = input(f"{question} (Y/n): ").casefold() or "y"
    return answer == "y"
```

For my tastes the one liners after that just become less readable (though still functional)

condar-metacell:
Looks all good to me at this point

ch-166.patch
alxbrd, line 1: This can be removed.

ddelpiano (approved):

PR #796, opened 2025-03-13
Enable experimental OPEN_API_3_1 support for schematesis; Update openapi.json look up logic

afonsobspinto:
Closes https://metacell.atlassian.net/browse/CH-178

# Implemented solution

- Enables experimental OPEN_API_3_1 support for schematesis;
- Looks up for openapi.json at / and /api

# How to test this PR

✦ ❯ harness-test . -i <application> -a
...

# Sanity checks:

- [x] The changes included in this pr are out of the current test coverage scope

- [x] The changes included in this PR are out of the current documentation scope

### DIFF
--- a/tools/cloudharness-test/cloudharness_test/apitest_init.py
+++ b/tools/cloudharness-test/cloudharness_test/apitest_init.py
@@ -1,8 +1,8 @@
 import os
 import logging
-import requests
 
 import schemathesis as st
+st.experimental.OPEN_API_3_1.enable()
 
 from cloudharness.auth import get_token
 
@@ -20,16 +20,18 @@ if "APP_URL" or "APP_SCHEMA_FILE" in os.environ:
             logging.exception("The local schema file %s cannot be loaded. Attempting loading from URL", openapi_uri)
 
     if not schema:
-        # remote testing: might be /api/openapi.json or /openapi.json
+        # Try app_url/openapi.json
         try:
-            openapi_uri = openapi_uri = app_url + "/openapi.json"
+            openapi_uri = app_url.rstrip("/") + "/openapi.json"
             logging.info("Using openapi spec at %s", openapi_uri)
             schema = st.from_uri(openapi_uri)
         except st.exceptions.SchemaError:
-            # Use alternative configuration
+            logging.warning("Failed to load schema from %s", openapi_uri)
+
+            # Then try app_url/api/openapi.json
             try:
-                openapi_uri = app_url.replace("/api", "") + "/openapi.json"
-                print(requests.get(openapi_uri))
+                openapi_uri = app_url.rstrip("/") + "/api/openapi.json"
+                logging.info("Using openapi spec at %s", openapi_uri)
                 schema = st.from_uri(openapi_uri)
             except st.exceptions.SchemaError as e:
                 raise Exception(


tools/cloudharness-test/cloudharness_test/apitest_init.py
```diff
@@ -1,43 +1,48 @@
 import os
 import logging
-import requests
-
 import schemathesis as st
+from schemathesis.hooks import HookContext
 
 from cloudharness.auth import get_token
+st.experimental.OPEN_API_3_1.enable()
```
Stranger6667, line 7: Just noted the double call which is redundant

tools/cloudharness-test/cloudharness_test/apitest_init.py
```diff
@@ -66,3 +71,24 @@ def set(self, case, data, context):
...
+
+            param_str = str(param)
+
+            # Reject if any unsafe substring is present.
+            if any(unsafe in param_str for unsafe in UNSAFE_VALUES):
+                return False
+
+            return True
+
+        return strategy.filter(valid_param)
```
Stranger6667, line 94: it also could be done with `filter_path_parameters` which reduces some boilerplate

```python
@st.hook
def filter_path_parameters(context: HookContext, x):
    # Extract the candidate value.
    param = x["key"] if isinstance(x, dict) and "key" in x else x
    
    if param is None or param == "":
        return True

    param_str = str(param)

    # Reject if any unsafe substring is present.
    if any(unsafe in param_str for unsafe in UNSAFE_VALUES):
        return False

    return True

File: tools/cloudharness-test/cloudharness_test/apitest_init.py [hunk 3]
```diff
@@ -66,3 +71,24 @@ def set(self, case, data, context):
                 case.headers = case.headers or {}
                 case.headers["Authorization"] = f"Bearer {data}"
                 case.headers["Cookie"] = f"kc-access={data}"
+
+    UNSAFE_VALUES = ("%")
```
Stranger6667, line 75: Should it be `("%", )` instead? Right now it will work the same as it is a single char, though for future-proofing it might be better to adjust it

tools/cloudharness-test/cloudharness_test/apitest_init.py
afonsobspinto: Thanks. Could you point me to any docs / code that explain when does the filter_path_parameters happen?

I learned about the before_generate_path_parameters [here](https://schemathesis.readthedocs.io/en/stable/_modules/schemathesis/hooks.html) but I don't see the filter_path_parameters there

afonsobspinto:
Thanks for the review @Stranger6667 👍

tools/cloudharness-test/cloudharness_test/apitest_init.py
Stranger6667: Yep, the docs are [here](https://schemathesis.readthedocs.io/en/stable/extending.html#hooks) (and a bit more [here](https://schemathesis.readthedocs.io/en/stable/extending.html#filtering-data)). Let me know if they aren't sufficient, but the original idea was to add shortcuts for that nested functions + `strategy.filter` calls

afonsobspinto: It's clear now, thanks. [This ](https://schemathesis.readthedocs.io/en/stable/extending.html#hook-locations) also helped.

PR #814, opened 2025-07-10
CH-21 fix JupyerHub dependency on accounts

iamludal:
https://metacell.atlassian.net/browse/CH-21


iamludal:
Closes https://github.com/MetaCell/cloud-harness/issues/414

PR #830, opened 2026-01-21
Changes due to IFN upgrade

ddelpiano:


zsinnema (commented):
@ddelpiano I think this should go into the development branch of CH, wdyt?

ddelpiano:
@zsinnema @filippomc what would you like me to do with this branch? I can see there are some changes that are reverting to older versions we currently have in develop, not sure this has to into develop or we want to bring it first in a release that is not that far from the upgrades that Zoran started and then merge all back to develop, let me know what you prefer. Thanks!

filippomc:
Better rebase on develop with only the commits that are necessary. Using the old library versions that I see in the PR is not advisable

ddelpiano:
closed in favour of https://github.com/MetaCell/cloud-harness/pull/847

PR #838, opened 2026-02-27
CH-243: Make route pattern customizable

iamludal:
Closes [CH-243](https://metacell.atlassian.net/browse/CH-243)

# Implemented solution

Make the route pattern customizable on the app level, as well as globally.

# How to test this PR

Change the route pattern of an app and try to reach it

# Sanity checks:
- [ x] All the linked issues are assigned

[CH-243]: https://metacell.atlassian.net/browse/CH-243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/deployment-configuration/helm/templates/ingress.yaml
+++ b/deployment-configuration/helm/templates/ingress.yaml
@@ -70,11 +70,12 @@ kind: Ingress
 metadata:
   name: {{ .Values.ingress.name | quote }}
   annotations:
-    kubernetes.io/ingress.class: nginx # Deprecated by Kubernetes, however still required for GKE
+    kubernetes.io/ingress.class: {{ .Values.ingress.ingressClass }} # Deprecated by Kubernetes, however still required for GKE
     {{- if and (not .Values.local) $tls }}
     kubernetes.io/tls-acme: 'true'
     cert-manager.io/issuer: {{ printf "%s-%s" "letsencrypt" .Values.namespace }}
     {{- end }}
+    {{- if eq .Values.ingress.ingressClass "nginx" }}
     nginx.ingress.kubernetes.io/ssl-redirect: {{ (and $tls .Values.ingress.ssl_redirect) | quote }}
     nginx.ingress.kubernetes.io/proxy-body-size: '{{ .Values.proxy.payload.max }}m'
     nginx.ingress.kubernetes.io/proxy-buffer-size: '128k'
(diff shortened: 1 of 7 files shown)


deployment-configuration/helm/templates/ingress.yaml
```diff
@@ -39,7 +40,7 @@
     {{- end }}
   {{- end }}
 {{- end }}
-      - path: /(.*)
+      - path: /{{ $routePattern }}
```
filippomc: better do `| quote`

zsinnema (changes requested):
**question**: is it correct that I'm missing a backwards compatible fallback value for the route_pattern value? I mean incase the application doesn't set any imho it should default to "/"

**suggestion**: I would go for the Traefik regular expression as default format/setting (K8s ingress nginx controller is deprecated).

iamludal:
> **question**: is it correct that I'm missing a backwards compatible fallback value for the route_pattern value? I mean incase the application doesn't set any imho it should default to "/"
> 
> **suggestion**: I would go for the Traefik regular expression as default format/setting (K8s ingress nginx controller is deprecated).

Regarding your question, I think it should be backward compatible as I added a `route_pattern` field in the default values file of the chart, with a default value of `/(.*)` (the current implementation). If you think I didn't put it in the correct file please let me know, I can also add a second fallback value to be very sure.

As for Traefik, do you mean they handle regex differently, or do you mean to use the `PathRegex` syntax instead of the current one?

zsinnema:
> > **question**: is it correct that I'm missing a backwards compatible fallback value for the route_pattern value? I mean incase the application doesn't set any imho it should default to "/"
> > **suggestion**: I would go for the Traefik regular expression as default format/setting (K8s ingress nginx controller is deprecated).
> 
> Regarding your question, I think it should be backward compatible as I added a `route_pattern` field in the default values file of the chart, with a default value of `/(.*)` (the current implementation). If you think I didn't put it in the correct file please let me know, I can also add a second fallback value to be very sure.
> 
> As for Traefik, do you mean they handle regex differently, or do you mean to use the `PathRegex` syntax instead of the current one?

Ah yes, you are right about 1. That's indeed the default fallback
About 2. yes I think it's a different format, not sure. May be you can have a look how they handle regex in path routes

iamludal:
> > > **question**: is it correct that I'm missing a backwards compatible fallback value for the route_pattern value? I mean incase the application doesn't set any imho it should default to "/"
> > > **suggestion**: I would go for the Traefik regular expression as default format/setting (K8s ingress nginx controller is deprecated).
> > 
> > 
> > Regarding your question, I think it should be backward compatible as I added a `route_pattern` field in the default values file of the chart, with a default value of `/(.*)` (the current implementation). If you think I didn't put it in the correct file please let me know, I can also add a second fallback value to be very sure.
> > As for Traefik, do you mean they handle regex differently, or do you mean to use the `PathRegex` syntax instead of the current one?
> 
> Ah yes, you are right about 1. That's indeed the default fallback About 2. yes I think it's a different format, not sure. May be you can have a look how they handle regex in path routes

It seems like regex are handled as regular regex using the `PathRegexp` syntax, see https://doc.traefik.io/traefik/v3.0/routing/routers/#path-pathprefix-and-pathregexp.

So I changed the manifest itself to be compatible with the Traefik ingress controller by adding the `traefik.ingress.kubernetes.io/router.pathmatcher: PathRegexp` annotation. 

See: https://doc.traefik.io/traefik/reference/routing-configuration/kubernetes/ingress/#on-ingress

iamludal:
[This Traefik doc](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/) says that Regex are using [this lib](https://pkg.go.dev/regexp), so Regex should follow the syntax accepted by [RE2](https://github.com/google/re2/wiki/Syntax).
I need to update the doc to make the exclusion example working with this syntax.

PR #840, opened 2026-03-09
Feature/gateway updates

filippomc:
---
Closes [CH-243](https://metacell.atlassian.net/browse/CH-243) - Customizable Ingress entry points. Took over from https://github.com/MetaCell/cloud-harness/pull/838, with some fixes and default for cross compatible configuration (Prefix/Exact) and support for Traefik.

---
Closes [CH-224](https://metacell.atlassian.net/browse/CH-224) - Gatekeeper update. Took over from https://github.com/MetaCell/cloud-harness/pull/839, handles the encryption key on a auto generated secret

---
Closes [CH-248](https://metacell.atlassian.net/browse/CH-248) - external database support. Adds a new parameter 
`harness.external_connect_string` on applications, which points the database route there. Uses a secret to store the connect string.

# How to test this PR

Currently deployed at https://samples.test-ch.dev.metacell.us/

# Sanity checks:

[CH-243]: https://metacell.atlassian.net/browse/CH-243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CH-224]: https://metacell.atlassian.net/browse/CH-224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CH-248]: https://metacell.atlassian.net/browse/CH-248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zsinnema (changes requested):
**comment**: @filippomc wdyt of always create the database connect string secret? I mean insterad of the current database connection method we could create the secret when installing the chart when it's not configured in the values.yaml of the microservice. Imho this is cleaner than having 2 different ways of getting the correct database connection in the service

filippomc:
> **comment**: @filippomc wdyt of always create the database connect string secret? I mean insterad of the current database connection method we could create the secret when installing the chart when it's not configured in the values.yaml of the microservice. Imho this is cleaner than having 2 different ways of getting the correct database connection in the service

@zsinnema I like the secrets idea as a chance to get rid of open and configured database passwords and databases, but have to make it backwards compatible. 
I like less delegating the logic of composing the connect string to helm chart templates from Python, I think it adds complexity rather than simplify.

All considered I lean towards leaving as it is for this PR scope.

zsinnema:
> > **comment**: @filippomc wdyt of always create the database connect string secret? I mean insterad of the current database connection method we could create the secret when installing the chart when it's not configured in the values.yaml of the microservice. Imho this is cleaner than having 2 different ways of getting the correct database connection in the service
> 
> @zsinnema I like the secrets idea as a chance to get rid of open and configured database passwords and databases, but have to make it backwards compatible. I like less delegating the logic of composing the connect string to helm chart templates from Python, I think it adds complexity rather than simplify.
> 
> All considered I lean towards leaving as it is for this PR scope.

@filippomc yeah I understand and share your concers.

What about doing this:
* removing the "external_" part of the secret name and configuration (values.yaml) 
* create the secret always for use in new services
* don't touch the current config (for existing services)

wdyt?

filippomc:
> * removing the "external_" part of the secret name and configuration (values.yaml)
> * create the secret always for use in new services
> * don't touch the current config (for existing services)

@zoran-sinnema your suggestion is introducing a new scope to the initial implementation content, which is only about giving a new option to use an external database. Will do the rename within this scope to make it future proof, but let's open another card for the rest. It's additional effort that can be prioritized separately.

zsinnema:
> > * removing the "external_" part of the secret name and configuration (values.yaml)
> > * create the secret always for use in new services
> > * don't touch the current config (for existing services)
> 
> @zoran-sinnema your suggestion is introducing a new scope to the initial implementation content, which is only about giving a new option to use an external database. Will do the rename within this scope to make it future proof, but let's open another card for the rest. It's additional effort that can be prioritized separately.

my only concern that remains for now is that this is a global "system" setting and not a per (micro)service setting.
I would have liked to see that it can be configured per service.

this is the follow up ticket: https://metacell.atlassian.net/browse/CH-249

filippomc:
> my only concern that remains for now is that this is a global "system" setting and not a per (micro)service setting. I would have liked to see that it can be configured per service.

This implementation is per microservice already, but of course could set value-template to be inherited by all microservices.

zsinnema (approved):

PR #842, opened 2026-03-30
Extra containers definition support: init and sidecars

filippomc:
Closes [CH-85](https://metacell.atlassian.net/browse/CH-85)

# Implemented solution

Added init and sidecar containers as part of the deployment (extraContainers)

Example:

```yaml
harness:
  ...
  deployment:
    ...
    extraContainers:
      init-print:
        auto: true
        initContainer: true
        commands: ["sh", "-c", "echo Initializing"]
        shareVolume: false
      sidecar-print:
        auto: true
        initContainer: false
        commands: ["sh", "-c", "while true; do echo Sidecar running at $(date); sleep 60; done"]
        shareVolume: false
```
# How to test this PR

Implemented as part of samples application. Can check the online deployment or deploy it locally and check the pods.

# Sanity checks:

[CH-85]: https://metacell.atlassian.net/browse/CH-85?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


application-templates/django-base/deploy/values.yaml
```diff
@@ -10,6 +10,28 @@ harness:
     - cloudharness-django
   use_services:
     - name: common
+  deployment:
```
filippomc, line 13: default django migration

application-templates/django-base/deploy/values.yaml
```diff
@@ -10,6 +10,28 @@ harness:
...
+    extraContainers:
+      run-migrations:
+        auto: true
+        initContainer: true
+        commands:
+          - sh
+          - -c
+          - |
+            python manage.py migrate --check 2>/dev/null && echo "No migrations needed" && exit 0
+            if python manage.py dbshell -c "SELECT pg_try_advisory_lock(42)" | grep -q t; then
```
zsinnema: **critical**: @filippomc this is not going to work because the lock is released after the dbshell ended. It should be atomic with the migration. It should be something like this
```
# management/commands/locked_migrate.py
from django.core.management.commands.migrate import Command as MigrateCommand
from django.db import connection

class Command(MigrateCommand):
    def handle(self, *args, **options):
        with connection.cursor() as cursor:
            cursor.execute("SELECT pg_advisory_lock(42);")
            super().handle(*args, **options)
```
and then execute `python manage.py locked_migrate`

zsinnema (changes requested):

zsinnema (approved):

PR #849, opened 2026-04-21
Fix thread safety and cache invalidation

zsinnema:
# Implemented solution

Use contextvars and invalidate user cache (on delete/update user event from KC)
Also added a README about how this Django app works and how to use it

- [x] The changes included in this PR are out of the current documentation scope

### DIFF
--- a/infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
+++ b/infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
@@ -1,0 +1,322 @@
+# cloudharness-django
+
+A Django library that integrates [CloudHarness](https://github.com/MetaCell/cloud-harness) infrastructure into Django applications. It handles Keycloak OIDC authentication, automatically synchronises Keycloak users/groups/organisations into Django's database, and listens to Kafka events to keep that data up to date in real time.
+
+## Table of contents
+
+- [How it works](#how-it-works)
+- [Installation](#installation)
+- [Configuration](#configuration)
+- [Startup](#startup)
+- [Models](#models)
+- [Authentication](#authentication)
+- [Services](#services)
+- [Authorization levels](#authorization-levels)
+- [Event-driven synchronisation](#event-driven-synchronisation)
+- [Management commands](#management-commands)
+- [Admin UI](#admin-ui)
+- [Local development](#local-development)
+
+---
+
+## How it works
+
+```
+HTTP request
+     │
+     ▼
+BearerTokenMiddleware
+     │  extracts JWT from Authorization header or kc-access cookie
+     │  decodes token → gets Keycloak user ID (sub)
+     │  checks Django cache (TTL 60 s)
+     │        hit  ──────────────────────────────────► request.user
+     │        miss ──► _get_user()
+     │                      │
+     │                      ├─ User.objects.get(member__kc_id=…)
+     │                      │       found  ─► sync_kc_user() (updates fields)
+     │                      │       not found ─► sync_kc_user() (creates User + Member)
+     │                      │
+     │                      └─► SAFETY CHECK: User must always have a Member
+     │                               ok  ─► cache user ─► request.user
+     │                               fail ─► keep anonymous
+     ▼
+Django view / Django Ninja
+
+
+Keycloak admin events (via Kafka)
+     │
+     ▼
+KeycloakMessageService.event_handler()
+     │
+     ├─ USER          ─► invalidate_user_cache() + sync_kc_user()
+     ├─ GROUP         ─► sync_kc_group()
+     ├─ CLIENT_ROLE_MAPPING ─► sync_kc_user()
+     ├─ GROUP_MEMBERSHIP    ─► sync_kc_user() (groups re-synced inside)
+     └─ ORGANIZATION_MEMBERSHIP ─► sync_kc_user() (orgs re-synced inside)
+```
+
+Every authenticated request syncs the Keycloak user into the Django `User` table. A `Member` record (holding the Keycloak UUID `kc_id`) is always created alongside it. The middleware caches the resolved `User` object for 60 seconds to avoid a database round-trip on every request. Kafka events invalidate that cache and trigger a full re-sync when Keycloak data changes.
+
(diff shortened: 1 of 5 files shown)


infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+```bash
+python manage.py migrate
+```
+
+---
+
+## Configuration
+
+Add the following to your `settings.py` after importing the CloudHarness settings:
```
filippomc, line 94: I'd mention that the preferred way to get started is to use `harness-application`

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+Every authenticated request syncs the Keycloak user into the Django `User` table. A `Member` record (holding the Keycloak UUID `kc_id`) is always created alongside it. The middleware caches the resolved `User` object for 60 seconds to avoid a database round-trip on every request. Kafka events invalidate that cache and trigger a full re-sync when Keycloak data changes.
+
+---
+
+## Installation
+
+```bash
+pip install cloudharness-django
```
filippomc, line 65: This might confuse and make the reader expect that cloudharness-django is on pip

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+Add the library's settings module to the **top** of your `settings.py`:
+
+```python
+PROJECT_NAME = "my-app"   # must be set before the import
```
filippomc, line 71: Not an issue with this specific PR, but having a name that's unrelated from the cloudharness configuration for the application can be confusing. `get_current_configuration().name` would be more consistent IMO

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+- Appends `cloudharness_django`, `admin_extra_buttons`, and `django_prometheus` to `INSTALLED_APPS`.
+- Inserts `BearerTokenMiddleware` (and Prometheus middleware) into `MIDDLEWARE`.
+- Configures the `DATABASES` setting from the CloudHarness deployment values (PostgreSQL in Kubernetes, SQLite on a developer machine when no `allvalues.yaml` is found).
+- Sets `CSRF_TRUSTED_ORIGINS` from the CloudHarness domain.
+- Redirects `ROOT_URLCONF` through `cloudharness_django.urls` (which mounts Prometheus metrics and then delegates to your original URL config).
+
+Run migrations once after installation:
+
+```bash
+python manage.py migrate
```
filippomc, line 87: We should probably suggest the preferred way to run migrations (the init container with the lock, now in templates)

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+        from cloudharness_django.services.events import init_listener_in_background
+
+        # Initialise auth + user services (creates the Keycloak client and roles if absent)
+        init_services()
+
+        # Start the Kafka consumer in a background thread
+        init_listener_in_background()
+```
+
+If Keycloak may not be reachable immediately (e.g. during container startup), use the background variant for services as well:
```
filippomc, line 144: as well -> instead?

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+2. Decodes and validates the JWT (via `cloudharness.auth.keycloak.AuthClient`).
+3. Looks up the Django `User` by `member__kc_id`, creating it if absent.
+4. Syncs user attributes and group memberships from Keycloak.
+5. Caches the result and sets `request.user`.
+
+If the token is invalid, the user is logged out and the `kc-access` cookie is deleted.
+
+### Cache invalidation
+
+When a user is deleted or deactivated in Keycloak the Kafka event handler calls `invalidate_user_cache()` to remove the stale cache entry before syncing. You can also call it directly:
```
filippomc, line 192: Is this relevant for the user of the library or more interesting for the library developer? Also this is the first time the kafka event handler is mentioned

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+if auth_level == AuthorizationLevel.ADMIN:
+    ...
+```
+
+---
+
+## Event-driven synchronisation
+
+`KeycloakMessageService` consumes the `keycloak.fct.admin` Kafka topic. On each Keycloak admin event it dispatches to the appropriate sync method:
```
filippomc, line 269: This deserves more context. The event based sync adds real time update of user properties, but it's not necessary for the users sync functionality to work in general

infrastructure/common-images/cloudharness-django/libraries/cloudharness-django/README.md
```diff
@@ -0,0 +1,322 @@
...
+
+    def ready(self):
+        from cloudharness_django.services import init_services
+        from cloudharness_django.services.events import init_listener_in_background
+
+        # Initialise auth + user services (creates the Keycloak client and roles if absent)
+        init_services()
+
+        # Start the Kafka consumer in a background thread
+        init_listener_in_background()
```
filippomc, line 141: Optional, and requires `harness.dependencies.soft += events`

filippomc (commented):
The scope of this PR is unclear. I suspect it's a fix for cached users within kafka listener workflows but not so sure, as I'd expect a fix for Kafka listeners to circumvent cache by construction instead of invalidating the cache.

filippomc (approved):

PR #852, opened 2026-05-01
CH-268: Fix name based exclusion in codefresh.py

alxbrd:
Improve image removal pruning logic when excluding an image with deps used by other images.

Closes [CH-268](https://metacell.atlassian.net/browse/CH-268)

[CH-268]: https://metacell.atlassian.net/browse/CH-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tools/deployment-cli-tools/ch_cli_tools/codefresh.py
+++ b/tools/deployment-cli-tools/ch_cli_tools/codefresh.py
@@ -40,6 +40,24 @@ def clean_step_key(s: str) -> str:
     return re.sub(r'[^a-zA-Z0-9_]', '_', s)
 
 
+def dockerfile_selector_candidates(base_path: str, dockerfile_path: str) -> set[str]:
+    """Return stable identifiers that may be used with include/exclude selectors."""
+    relative_to_base = get_app_relative_to_base_path(base_path, dockerfile_path)
+    parent_name = relative_to_base.split("/")[0] if relative_to_base else ""
+    return {
+        candidate for candidate in (
+            relative_to_base,
+            app_name_from_path(relative_to_base),
+            parent_name,
+        ) if candidate
+    }
+
+
+def path_contains_excluded_segment(path: str, excluded_segments) -> bool:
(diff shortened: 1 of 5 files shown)


tools/deployment-cli-tools/ch_cli_tools/codefresh.py
```diff
@@ -40,6 +40,24 @@ def clean_step_key(s: str) -> str:
     return re.sub(r'[^a-zA-Z0-9_]', '_', s)
 
 
+def dockerfile_selector_candidates(base_path: str, dockerfile_path: str) -> set[str]:
+    """Return stable identifiers that may be used with include/exclude selectors."""
+    relative_to_base = get_app_relative_to_base_path(base_path, dockerfile_path)
+    parent_name = relative_to_base.split("/")[0] if relative_to_base else ""
```
aranega, line 46: Should we split on  `/` or on `os.path.sep`? I'm not sure what's expected as path, in the function `path_contains_excluded_segment` it seems that a path could contain os dependent separator, so I'm not sure.

aranega (commented):
Thanks for the PR @alxbrd ! It looks good to me. I'm only not sure about the fact that we consider unix-path separator in some functions, but perhaps it's an assumption that is done on the whole base code.

tools/deployment-cli-tools/ch_cli_tools/codefresh.py
alxbrd: Thanks @aranega, that is a good question. Looking at other utils functions, the code assumes "/" rather than system path separator.

alxbrd: Created this task to fix this across the whole app: https://metacell.atlassian.net/browse/CH-269

filippomc (changes requested):
Tests are failing, let's fix before merging

filippomc (approved):

PR #861, opened 2026-08-19
Configure base images

aranega:
Closes [CH-287](https://metacell.atlassian.net/browse/CH-287)

# Implemented solution

The solution relies on the addition of an extra entry in the `value-template.yaml` file which allows the user to declare a mapping to change the base image. The mapping affects the build args which are passed to skaffold to build the images, effectively replacing the base image with the one declared in the `value-template.yaml` file. The PR also document about the path to override for apps wich do not rely on an accessible `Dockerfile` (e.g: argo, or the gatekeeper) 

Note: Some tests realted to secure secret are failing on `develop` the same pre-existing failing tests are still failing with this PR.

# How to test this PR

To test the PR, the `value-template.yaml` file can be updated with a specific mapping for the source image, e.g:

```yaml
source_images:
  MY_BASE_IMAGE: "mybaseimg:123"
```
and the new deployment configuration generated using `harness-deployment`. 
After the generation, the `skaffold.yaml` file can be opened to see that `buildArgs` have been impacted for `myapp` and that the new base image `mybaseimg:123` have been properly referenced.

# Sanity checks:
- [X] The pull request is explicitly linked to the relevant issue(s)
- [X] The issue is well described: clearly states the problem and the general proposed solution(s)
- [X] In this PR it is explicitly stated how to test the current change
- [X] The issue seta the scope and the type of issue (bug, story, task, etc.)
- [X] The relevant components are indicated in the issue (if any)
- [X] All the linked issues are included in one Sprint
- [X] All the linked issues are in the Review state
- [X] All the linked issues are assigned

- [X] The present changes do not change the preexisting api in any way

- [X] This PR and the issue are tagged as `alert:deployment`

- [X] Tests for the relevant cases are included in this pr

- [X] The documentation has been updated to match the current changes

[CH-287]: https://metacell.atlassian.net/browse/CH-287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/base-common-images.md
```diff
@@ -60,3 +60,78 @@ libraries or to provide new libraries to share within all applications.
...
+Changing the base image for your application or applications which have a `Dockerfile` is done through the `source-images` entry of your `value-template.yaml` file. 
+This entry defines a mapping between the `ARG` of your `Dockerfile` and the value you want to inject. 
+Here is an example of a declared mapping for two applications:
+
+```yaml
+# value-template.yaml
+image-sources:
+  myapp: "mybaseimg:14.5"
+  samples:
+    CLOUDHARNESS_FRONTEND_BUILD: "myfrontendbaseimg:15.6"
```
filippomc: bad example: CLOUDHARNESS_FRONTEND_BUILD is one we are building, no point in redefining at this level. Could be useful if we want to skip some base image build steps in some pipeline, but not sure if we really want this

filippomc (changes requested):
I'm afraid the configuration is more complicated that it needs to be, as this is supporting the reuse of the same name in different context, which is unnecessary.

I think it's better to use one name for each image that is used, and make it possible to override it, regardless of where it's used.

For instance, cloudharness-frontend-build needs a node image. I'd do `FROM $BASE_NODE` and then 

```yaml
image_sources:
  BASE_NODE: node: 22
  GATEKEEPER: quay.io/gogatekeeper: 5.0.0
  KEYCLOAK: quay.io/keycloak: 26.0.0
  ...
```
We can pass all args to all builds then

docs/base-common-images.md
```diff
@@ -60,3 +60,75 @@ libraries or to provide new libraries to share within all applications.
...
+
+### Change the base image for applications with a `Dockerfile`
+
+Changing the base image for your application or applications which have a `Dockerfile` is done through the `source_images` entry of your `value-template.yaml` file. 
+This entry defines a mapping between the `ARG` of your `Dockerfile` and the value you want to inject. 
+Here is an example of a declared mapping to change the base image for Python based apps and Node base apps:
+
+```yaml
+# value-template.yaml
+image_sources:
```
filippomc: should be source_images?

filippomc (approved):

PR #866, opened 2026-09-03
Override non-built source images

filippomc:
Closes [CH-287](https://metacell.atlassian.net/browse/CH-287)

# Implemented solution

- Add support for sub chart images
- Add support for other hardcoded images (dbs, etc)
- Update docs

All image paths/values are exported as values-overrides.yaml, so it's easier to change everything in one place.
`harness-deployment` generate that file with every image that's relevant

# How to test this PR

Run `harness-deployment` and then check `deployment/helm/values-overrides.yaml`.
It should contain all the relevant source images.

# Sanity checks:
- [ ] The issue seta the scope and the type of issue (bug, story, task, etc.)

[CH-287]: https://metacell.atlassian.net/browse/CH-287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/deployment-cli-tools/ch_cli_tools/codefresh.py
```diff
@@ -473,6 +473,11 @@ def adjust_build_steps(index):
     if deployment_step:
         arguments = deployment_step.get("arguments")
         if arguments:
+            # Apply the generated image overrides after values.yaml so editing them wins
+            values_overrides_file = f"./{DEPLOYMENT_PATH}/{HELM_CHART_PATH}/{VALUES_OVERRIDES_PATH}"
```
aranega: nitpick: it could be cool to use more `Path` object from the `pathlib` builtin libs as it removes the use of `/` as os dependent separator. This is something which is not entirely coherent through our base code (CH-269 points to this as something to solve), but in some cases, I'm not sure if that's meant to be run only on the CI/CD (usually linux-based) or on the user machine.

aranega (approved):
The PR looks good to me, I just added a comment/nitpick about the fact of using some hardcoded path, but I'm unsure the line in question is something which is supposed to run on the CI/CD or on the user's machine.

tools/deployment-cli-tools/ch_cli_tools/codefresh.py
filippomc: It's a good argument to attempt to harmonize the way we handle paths, but as long as the code is not fully coherent it's a hard call to enforce it on a single PR. What about a PR that targets paths specifically and moves everything to pathlib?

PR #873, opened 2026-09-16
CH-293 support for application multi  instances

filippomc:
Closes [CH-293](https://metacell.atlassian.net/browse/CH-293)

# Implemented solution

An application can be deployed several times from the same build by declaring instances as directories under its `deploy/instances`. Each instance directory holds a `values.yaml` with the values it changes and, optionally, `resources` and `templates` overlaid on the application's.

An instance is deployed as the application `[application]-[instance]` (e.g. `samples-instance1`), on its own subdomain (the instance directory name by default), with its own service, deployment, database, volume claim, secrets and gatekeeper. It runs the image built for the application and inherits its whole configuration. Values are layered `instances/x/values-[ENV].yaml` > `instances/x/values.yaml` > `deploy/values-[ENV].yaml` > `deploy/values.yaml`. What identifies the application (`subdomain`, `aliases`, `domain`, resource names, volume claim name, `database.connect_string`) is not inherited.

`-i app` includes the application's instances, `-i app-instance` pulls in the application, `-e app-instance` excludes one instance. An instance whose key collides with an application or a task image is rejected with an explicit error.

Instances are handled in the new `ch_cli_tools/instances.py` and derived while their application is read, so the generators treat them as ordinary applications from there on. The `deploy/instances` directories are the only record of what is an instance; the configuration schema is unchanged. Also included: the Dockerfile filter in `utils.is_buildable_dockerfile_path` matches path segments instead of substrings, and `_secrets.tpl` separates external secret manager resources with `---` after each resource. Documentation in `docs/applications/README.md`.

# How to test this PR

- Run `pytest tests/test_helm.py tests/test_dockercompose.py` in `tools/deployment-cli-tools` within the `ch` conda environment; the `test_instance*` tests cover the feature.
- Run `harness-deployment . -i samples -d my.local -e dev` and check that `deployment/helm/values.yaml` contains `apps.samples-instance1` with `subdomain: samples1`, `build: false` and the image of `samples`, and that `deployment/helm/resources/samples-instance1/` holds both `example.yaml` (from the instance) and `myConfig.json` (inherited).
- Run `helm template deployment/helm` and check the `samples-instance1` StatefulSet and Service, the ingress rule for `samples1.my.local` and the `samples1-gk` gatekeeper.

# Sanity checks:
- [x] The issue seta the scope and the type of issue (bug, story, task, etc.)
- [x] All the linked issues are included in one Sprint
- [x] All the linked issues are in the Review state

[CH-293]: https://metacell.atlassian.net/browse/CH-293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/applications/README.md
```diff
@@ -70,6 +70,77 @@ harness:
...
+        resources/
+          example.yaml         # overrides the application's resource of the same name
+        templates/             # optional, overlaid on the application's templates
+```
+
+The instance above is deployed as the application `samples-instance1`: that key names its service,
+deployment, database, volume, gatekeeper and configmaps, and is how it is referenced on the command
+line. It runs the image built for `samples` — an instance adds no build, so declare no Dockerfile
+in it. The key must not take over one of the application's task images: an instance named `print`
+on an application with a `tasks/print-file` is rejected, as `samples-print` would own
```
aranega: I'm not sure I understand the rejection example and why. It's because `print-file` would be defined in the `samples-print` instance? The example is related to `samples` then? I think I don't get what `would own` means in this context, or where `tasks/print-file` is supposed to be. Is it declared for `samples` or `samples-print`?

docs/applications/README.md
```diff
@@ -70,6 +70,77 @@ harness:
...
+  `subdomain: null` to give an instance no ingress at all
+- the names of the service, deployment and database, which are derived from the instance key
+- `deployment.volume.name`, prefixed with the instance key, so the instance never mounts the
+  application's storage
+- `database.connect_string`, emptied: an instance of an application using an externally managed
+  database needs a connection string of its own. Set `database.auto: true` to have CloudHarness
+  deploy a database of its own for it instead.
+
+Instances are deployed together with their application: `harness-deployment -i samples` deploys
+`samples` and all its instances, and `-e samples-instance1` leaves one out. CI builds and tests the
```
aranega: is it `-e` or `-ex` to remove the deployment of the instance?

aranega (changes requested):
Just a couple of comments on the readme, all the rest looks good to me :)

docs/applications/README.md
filippomc: good point, doesn't sound like a clash to me too

